PR #23, opened 2019-06-24
Add Roster replacement

tharvik:
* `config.ts` contains `LoadConfig`, as way to retrieve the initial list of nodes from the server, and `Config`, a trivial information container
* `connections.ts` contains `IConnection`, the interface for a connection, `WebSocketConnection` implement it with websocket, and `MultiConnections`, a dispatcher for multiple connections.

### DIFF
--- a/dynacred/src/config.ts
+++ b/dynacred/src/config.ts
@@ -1,24 +1,25 @@
-import * as toml from 'toml';
+import * as toml from "toml";
 
 /**
  * Readonly container for the configuration
  *
  * TODO preciser name
-*/
+ */
 export class Config {
     constructor(
         readonly addresses: Addresses,
     ) {}
 
     /**
-    * Type containing both the addresses of the endpoints for servers and for
-    * clients
-    */
+     * Type containing both the addresses of the endpoints for servers and for
+     * clients
+     */
+    // tslint:disable-next-line
     private class Addresses {
         constructor(
             readonly servers: string[],
             readonly clients: string[],
-        ) {}
+        ) {} // tslint:disable-line
     }
 }
 
@@ -26,36 +27,43 @@ export class Config {
  * Load the Config from the server
  */
 export default async function LoadConfig(): Promise<Config> {
-    const res = await fetch(window.location.origin + '/assets/conodes.toml');
-    if (!res.ok)
+    const res = await fetch(window.location.origin + "/assets/conodes.toml");
+    if (!res.ok) {
         return Promise.reject(new Error(`on fetch config: ${res.status}: ${res.body}`));
+    }
 
     const parsed = toml.parse(await res.text());
-    if (!('servers' in parsed))
+    if (!("servers" in parsed)) {
         return Promise.reject(new Error(`on parse config: unable to find "servers"`));
-    if (!(Symbol.iterator in parsed.servers))
+    }
+    if (!(Symbol.iterator in parsed.servers)) {
         return Promise.reject(new Error(`on parse config: "servers" is not iterable`));
+    }
 
-    let addrs_client:string[] = [], addrs_server:string[] = [];
+    const addrsClient: string[] = [];
+    const addrsServer: string[] = [];
     for (const s of parsed.servers) {
-        if (!('Address' in s))
+        if (!("Address" in s)) {
             return Promise.reject(new Error(`on parse config: unable to find "Address" in each server`));
+        }
 
         const addr = s.Address;
-        if (typeof addr != 'string')
+        if (typeof addr !== "string") {
             return Promise.reject(new Error(`on parse config: "Address" of server is not a string`));
+        }
 
-        const [scheme, host, port_str] = addr.split(':');
-        if (scheme != 'tls')
+        const [scheme, host, port] = addr.split(":");
+        if (scheme !== "tls") {
             return Promise.reject(new Error(`on parse config: "Address" of server doesn't start with "tls"`));
+        }
 
-        addrs_server.push(addr);
+        addrsServer.push(addr);
         // FIXME hack as in cothority
-        addrs_client.push(`ws:${host}:${parseInt(port_str) + 1}`);
+        addrsClient.push(`ws:${host}:${parseInt(port, 10) + 1}`);
     }
 
     return new Config(new Addresses(
-        addrs_server,
-        addrs_client,
+        addrsServer,
+        addrsClient,
     ));
 }
--- a/dynacred/src/config.ts
+++ b/dynacred/src/config.ts
@@ -1,0 +1,61 @@
+import * as toml from 'toml';
+
+/**
+ * Readonly container for the configuration
+ *
+ * TODO preciser name
+*/
+export class Config {
+    constructor(
+        readonly addresses: Addresses,
+    ) {}
+
+    /**
+    * Type containing both the addresses of the endpoints for servers and for
+    * clients
+    */
+    private class Addresses {
+        constructor(
+            readonly servers: string[],
+            readonly clients: string[],
+        ) {}
+    }
+}
+
+/**
+ * Load the Config from the server
+ */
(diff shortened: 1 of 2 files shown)


dynacred/src/config.ts
```diff
@@ -0,0 +1,61 @@
+import * as toml from 'toml';
+
+/**
+ * Readonly container for the configuration
+ *
+ * TODO preciser name
```
ineiti: Suggestions?

dynacred/src/config.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+}
+
+/**
+ * Load the Config from the server
+ */
+export default async function LoadConfig(): Promise<Config> {
+    const res = await fetch(window.location.origin + '/assets/conodes.toml');
+    if (!res.ok)
+        return Promise.reject(new Error(`on fetch config: ${res.status}: ${res.body}`));
```
ineiti: Does this pass the `tslint.json`?
1. Line 29 and others: single quotes instead of double quotes
2. `if` without the `{}`

dynacred/src/config.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const parsed = toml.parse(await res.text());
+    if (!('servers' in parsed))
+        return Promise.reject(new Error(`on parse config: unable to find "servers"`));
+    if (!(Symbol.iterator in parsed.servers))
+        return Promise.reject(new Error(`on parse config: "servers" is not iterable`));
+
+    let addrs_client:string[] = [], addrs_server:string[] = [];
```
ineiti: This is still `const` in javascript, even if you `push` into it...

dynacred/src/config.ts
```diff
@@ -0,0 +1,61 @@
...
+        if (typeof addr != 'string')
+            return Promise.reject(new Error(`on parse config: "Address" of server is not a string`));
+
+        const [scheme, host, port_str] = addr.split(':');
+        if (scheme != 'tls')
+            return Promise.reject(new Error(`on parse config: "Address" of server doesn't start with "tls"`));
+
+        addrs_server.push(addr);
+        // FIXME hack as in cothority
+        addrs_client.push(`ws:${host}:${parseInt(port_str) + 1}`);
```
ineiti: This won't work, and I'm sure it got replaced: It should search for the `URL` field and use this, or fall back to `+1`.

Also: the servers need their public keys of all the services to verify the proofs.

dynacred/src/config.ts
tharvik: Not really for now, depending on how much data we want to retrieve from the server: For now, the most precise name I can think of is `Conodes`.

dynacred/src/config.ts
tharvik: Did not, do now.

dynacred/src/config.ts
tharvik: Done (handled by tslint).

dynacred/src/config.ts
ineiti: +1 - I still remember my first braceless `if` that cost me 1 day of debugging. So everytime I see it again, I frown...

dynacred/src/config.ts
ineiti: What is wrong with `Config`? Looks good to me...

dynacred/src/config.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        const [scheme, host, port] = addr.split(":");
+        if (scheme !== "tls") {
+            return Promise.reject(new Error(`on parse config: "Address" of server doesn't start with "tls"`));
+        }
+
+        addrsServer.push(addr);
+        // FIXME hack as in cothority
+        addrsClient.push(`ws:${host}:${parseInt(port, 10) + 1}`);
```
ineiti: This won't work, and I'm sure it got replaced: It should search for the URL field and use this, or fall back to +1.

Also: the servers need their public keys of all the services to verify the proofs.

dynacred/src/config.ts
```diff
@@ -0,0 +1,76 @@
...
+            }
+            if (!value.startsWith("tls:")) {
+                return Promise.reject(new Error(`on parse config: "${key}" of server doesn't start with "tls:"`));
+            }
+
+            return Promise.resolve(value);
+        };
+
+        const addr = await getField("Address");
+        const url = await getField("Url").catch((err) => {
```
ineiti: I think the field-name is `URL`, all-caps.

dynacred/src/config.ts
tharvik: Now that I'm adding more elements to it, `Config` looks good enough.

dynacred/src/config.ts
tharvik: It seems to be `Url` (from https://github.com/dedis/cothority/blob/master/dedis-cothority.toml)

dynacred/src/config.ts
tharvik: Done, the public keys induced a revamp, as much of `Config` changed (dedicated `Server` type and co, reduce size of checks, ...)

dynacred/src/config.ts
ineiti: Oh no - it seems different version do different things :( Here is what I have on my server:

```
$ cat conode_data/public.toml 
[[servers]]
  Address = "tls://conode.c4dt.org:7770"
  URL = "https://conode.c4dt.org:7771"
  Suite = "Ed25519"
...
```

So it should accept both `Url` and `URL`.

ineiti: Found the following in the latest java-code:

```
                // Special case: allow URL when it should be Url
                if (srvToml.contains("URL")) {
                    srv.Url = srvToml.getString("URL");
                }
```

tharvik: So, we have a choice now, do we support ill-formatted configuration or not (I lean towards the latter), if not, which format do we take?

tharvik: After discussion with @ineiti: support both, correct format it `Url`.

tharvik: Done.

ineiti:
Can you explain why you're not using https://github.com/dedis/cothority/blob/master/external/js/cothority/src/network/proto.ts#L56 for getting the servers? This would create a correct `Roster` containing `ServerIdentity`s.

tharvik:
> Can you explain why you're not using https://github.com/dedis/cothority/blob/master/external/js/cothority/src/network/proto.ts#L56 for getting the servers? This would create a correct `Roster` containing `ServerIdentity`s.

If the idea was only to provide a way to set the TOML we wanted at runtime, then this PR would have been way smaller. The idea behind having a Roster replacement is to have it mostly as a simple data container, named-tuple style, without implementing ProtoBuf interfacing or weird seemingly equal calls (`ServerIdentity. {urlToWebsocket,addressToWebsocket}`). This way, instead of passing around a `Roster` instance and using its functions, one module would receive directly the URL to connect to with `string[]` (or a data container via `Server[]`), improving general isolation.

ineiti:
OK, I see. But you pay this _general isolation_ with the fact that now you need to track whatever changes might happen in the `Roster`. Also, for some methods you will need the whole `Roster` (like creating a new ByzCoin - rare, I now, but still possible).

Also, now we have two representations of a roster: one the _cothority_ representation, and then your simple representation. And I don't see how _general isolation_ is a killer feature for this double representation.

tharvik:
Why do I have to track every changes? The modules implementing simple-roster won't need to be updated when roster is changed. For the whole `Roster` object needed in ByzCoin, I've look at the uses there of it, and none seems more involving than inlining a few times, not that much of change.

Yeah, I agree, having these two representation is quite annoying, so we should choose one (of course, I prefer my version, but I'm kinda biased on the matter).

tharvik:
Closing in favoe of #28.

PR #28, opened 2019-07-01
dynacred: get Roster from server's asserts

tharvik:
- get the `Roster` configuration from `/assets/conodes.toml`
- because getting the `Roster` is a remote operation, it should be awaited now (as `Defaults` is not scoped, we can't `await` it before using it)


ineiti:
First C4DT `Ready4Merge` merge!

PR #32, opened 2019-07-02
webserver: add wordpress login

tharvik:
* add call `/api/v0/wordpress/login`
* for now, `darcID` isn't used, we only check that the user is on the chain


webserver/src/app/api/v0/wordpress/login/login.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
ineiti: `component.spec.ts` should be removed...

tharvik: Removed in #35

PR #42, opened 2019-07-10
Cleanup and local testing

ineiti:
Cleanup:
- renamed webserver to webapp
- removed conodes.toml from git, as this should be in the local installation only

Local testing:
- added Defaults.AssetsPath, so that conodes.toml can be found in local testing


webapp/package.json
```diff
@@ -49,6 +49,7 @@
     "@types/jasminewd2": "~2.0.3",
     "@types/node": "~8.9.4",
     "codelyzer": "~4.5.0",
+    "cross-fetch": "^3.0.4",
```
tharvik: It is only used in dynacred, not in webserver, so it shouldn't be in devDeps

ineiti: but if I do `make dynacred_src`, then I need it. But only in dev, not in production, as the latter doesn't (shouldn't) depend on `make dynacred_src`...

ineiti:
> removing conodes.toml means that we have to provide it on c4dtsrv before merging

Absolutely correct.

ineiti:
I'm not sure, however, if a copied-in `conodes.toml` would survive `git clean -xdv` from `update.sh`. According to my reading of `man git-clean`, it would disappear.

webapp/package.json
tharvik: Ha right, good for me then

tharvik:
> I'm not sure, however, if a copied-in `conodes.toml` would survive `git clean -xdf` from `update.sh`. According to my reading of `man git-clean`, it would disappear.

It would disappear; we can copy it after `git clean`, as we are already doing some srcs modifications

ineiti:
> removing conodes.toml means that we have to provide it on c4dtsrv before merging

Or we define the checked out directory as _clean_ and remove the `git clean` command ;) Both are fine for me.

tharvik:
> > removing conodes.toml means that we have to provide it on c4dtsrv before merging
> 
> Or we define the checked out directory as _clean_ and remove the `git clean` command ;) Both are fine for me.

Hum, I was thinking of moving `update.sh` inside the repo at some point, so that travis does nearly all the job, and as we don't want to bundle a `conodes.toml`, we would still need a deploy script on the server copying the wanted toml. So I'm leaning toward the first fix (copying it afterward).

ineiti:
OK, probably we still need to work on a more _real_ deployment scenario with the angular projects, where we can actually `npm i` it or something... using `git pull` and then `ng build` seems strange. What's the 'normal' angular way?

tharvik:
> OK, probably we still need to work on a more _real_ deployment scenario with the angular projects, where we can actually `npm i` it or something... using `git pull` and then `ng build` seems strange. What's the 'normal' angular way?

From discussion on matrix, it seems that using `ng build --prod` then copying the `dist` dir to server is the way to go.

tharvik:
`cleanup_local_testing-fw` contains a forwarded version.

ineiti:
Close in favour of #43

PR #48, opened 2019-07-15
test build --prod

ineiti:
For unknown reasons, 'ng build --prod' does a more thorough test with regard
to private/public attributes than 'ng build' only.


tharvik:
Don't forget to also bump the `bin/update.sh` script.

PR #50, opened 2019-07-16
Trim `Defaults`

tharvik:
Remove many elements of `Defaults`. The remaining will be put in `config.{toml,ts}`.

### DIFF
--- a/webapp/src/app/admin/contacts/contacts.component.ts
+++ b/webapp/src/app/admin/contacts/contacts.component.ts
@@ -64,11 +64,6 @@ export class ContactsComponent implements OnInit {
 
     async createContact(view?: string) {
         const creds: IUserCred = {alias: "", email: "", view: "default", groups: []};
-        if (Defaults.Testing) {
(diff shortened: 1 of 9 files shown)


webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -145,16 +139,7 @@ export class ContactsComponent implements OnInit {
                         await this.uData.save();
                         const url = this.location.prepareExternalUrl("/register?ephemeral=" +
                             newUser.keyIdentity._private.toHex());
-                        let host = window.location.host;
-                        if (Defaults.Testing) {
-                            if (host.match(/local[1-9]?/)) {
```
ineiti: I would very much like to keep this functionality: when in testing mode and the host is one of `local[1-9]?`, it should take the next `local` as hostname. This is a very simple but good enough way to have multiple identities on the same browser for testing...

tharvik: Yeah, I see the usefulness of this, I put it back.

PR #55, opened 2019-07-23
dynacred: trim {package,tsconfig}.json

tharvik:
* correctly expose dynacred as a module (export needed fields), so we deps directly on `@c4dt/dynacred`
* remove much unused dev dependencies
* remove `publish.sh` as we can now directly use `npm publish`
* deps on https://github.com/dedis/cothority/pull/2018 (and its release)


ineiti:
Wow - looks good... dynacred is getting nice ;)

PR #56, opened 2019-07-24
dynacred: add Config, rm Defaults

### LINKED ISSUE #38: remove auto-Defaults
Somewhat related to c4dt/services#5.

- [ ] allow to select the URL to get conodes.toml from argv
- [ ] `ngOnInit`'ll be used to get it when using GUI

tharvik:
* add `@c4dt/dynacred.Config`, which loads a `config.toml` (file format is `conodes.toml` with some added root fields), the way to access it is via a field of `Data`
* remove `@c4dt/dynacred.Defaults` (yeay, finally)
* auto fill UserData at init time

Closes #38


ineiti:
How would this work for the `mobile` platform, specifically for testing on the local computer? Would you have to set up a local webserver to serve the configuration files? Or is there a hook in `Data` that lest you directly inject the configuration?

tharvik:
> How would this work for the `mobile` platform, specifically for testing on the local computer?

Currently, there is this small http server spawned during tests, with it, we can provide mock/real call for any config file needed.

> Would you have to set up a local webserver to serve the configuration files?

With this PR, no, because there isn't tests requiring it.

> Or is there a hook in `Data` that lest you directly inject the configuration?

`Config` doesn't have any notion of networking, it would be nice to also make `Data` this way. The ~hook that one can use is its constructor.

dynacred/src/Config.ts
```diff
@@ -0,0 +1,52 @@
...
+        const getIDField = (name: string): ID => {
+            if (!(name in parsed)) {
+                throw Error(`field "${name}" not found in config`);
+            }
+            return parseIDField(name, parsed[name]);
+        };
+
+        return new Config(
+            getIDField("ByzCoinID"),
+            Roster.fromTOML(raw),
```
ineiti: I'm a little bit hesitant if it wouldn't be better to put the roster in it's own toml-table:

```
ByzCoinID = "..."

[roster]
  [[servers]]
    ...
  [[servers]]
```

Not sure if it's worth it. Any thoughts?

dynacred/src/Data.ts
```diff
@@ -297,7 +300,7 @@ export class Data {
      * fields for initialization of the class.
      * @param obj (optional) object with all fields for the class.
      */
-    constructor(obj: any = {}) {
+    constructor(readonly config: Config, obj: any = {}) {
```
ineiti: Giving the `Config` here doesn't really make sense. I can think only of the two following use-cases:
- `Data.load()` finds data in the browser -> config not needed
- `Data.connectByzcoin()` needs to setup `bc` -> pass config (or better, ByzCoinRPC) there

dynacred/src/Data.ts
```diff
@@ -357,7 +360,7 @@ export class Data {
     async connectByzcoin(): Promise<ByzCoinRPC> {
```
ineiti: This should either take a `ByzCoinRPC = null` or a `Config = null`. Personally I tend to put everything as soon as possible into the most interpreted way possible, so I tend to propose the following:
`async connectByzcoin(bc: ByzCoinRPC = null)`
and then
```
if (this.bc == null && bc == null){
  throw new Error("No ByzCoin definition found")
}
```

dynacred/src/Data.ts
```diff
@@ -388,7 +391,7 @@ export class Data {
 
         this.contact.data = this;
         await this.contact.updateOrConnect();
-        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, await Defaults.RosterCalypso);
+        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, this.config.roster);
```
ineiti:
```suggestion
        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, this.bc.getConfig().roster);
```

dynacred/src/Data.ts
```diff
@@ -846,7 +849,7 @@ export class Data {
         this.contact = await Contact.fromByzcoin(this.bc, CredentialInstance.credentialIID(pub.toBuffer()));
         this.contact.data = this;
         await this.contact.updateOrConnect(this.bc);
-        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, await Defaults.RosterCalypso);
+        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, this.config.roster);
```
ineiti:
```suggestion
        this.lts = new LongTermSecret(this.bc, this.contact.ltsID, this.contact.ltsX, this.bc.getConfig().roster);
```

webapp/src/app/admin/status/status.component.ts
```diff
@@ -36,8 +34,9 @@ export class StatusComponent implements OnInit {
 
   async update() {
     this.nodes = [];
-    const list = (await Defaults.Roster).list;
-    const srpc = new StatusRPC(await Defaults.Roster);
+    const roster = this.uData.config.roster;
```
ineiti:
```suggestion
    const roster = this.uData.bc.getConfig().roster;
```

webapp/src/app/api/v0/wordpress/login/login.component.ts
```diff
@@ -47,7 +46,7 @@ export class LoginComponent implements OnInit {
         let loginOK = false;
         try {
             const ident = IdentityWrapper.fromIdentity(this.uData.keyIdentitySigner);
-            const reply = await this.uData.bc.checkAuthorization(Defaults.ByzCoinID, this.darcID, ident);
+            const reply = await this.uData.bc.checkAuthorization(this.uData.config.byzCoinID, this.darcID, ident);
```
ineiti:
```suggestion
            const reply = await this.uData.bc.checkAuthorization(this.uData.bc.getGenesis().computeHash(), this.darcID, ident);
```

dynacred/src/Data.ts
```diff
@@ -222,7 +225,7 @@ export class Data {
         return DarcInstance.fromByzcoin(bc, signer[0]);
     }
 
-    static async attachDevice(url: string): Promise<Data> {
+    static async attachDevice(config: Config, url: string): Promise<Data> {
```
ineiti: same thing as in `connectByzcoin` - I prefer using the interpretet config here as `bc: ByzCoinRPC`.

dynacred/src/Config.ts
tharvik: I don't think it is worth it, the roster is already contained inside the `servers` key.

tharvik:
> Looks good - there is mostly one big change that I still like to be put in: remove the `Config` from `Data`, and only pass `ByzCoinRPC` around.

I'm all for moving `Config` away from `Data`, but using `ByzCoinRPC` instead is really weird for me: what's the link between the two, I see no connection between configuration and byzcoin? Also `Config` actually gives to `ByzCoinRPC` the roster, wouldn't be there a chicken-in-the-egg situation?

> This moves `Config` further to the edge of the app and out of `Data`.

Out of `Data`, yes, but not to the edge, rather inside of `ByzCoinRPC`, as it would be the way to retrieve it.

> Perhaps it can even be removed from `dynacred` and moved into `webapp`?

The roster part of the config is used by `Data` and by `PersonhoodRPC`, both contained inside `dynacred`, moving it outside would be weird.

ineiti:
Can you do the changes with regard to `this.bc.getConfig()`, so that it always uses the latest definition of the roster?

tharvik:
Now, `Data` have no knowledge of `Config`, only of `ByzCoinRPC` (which can be easily created by `Config`).

PR #58, opened 2019-08-06
remove view- and groups-choice from c4dt_partner createUser

ineiti:
The C4DT partners should not be able to chose the view and the groups from the user,
but this should be created automatically for them. If they _do_ want to influence
this, they have to use the /admin-interface.

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -24,7 +24,7 @@
     "@angular/platform-browser": "^7.2.15",
     "@angular/platform-browser-dynamic": "^7.2.15",
     "@angular/router": "^7.2.15",
-    "@c4dt/cothority": "^3.1.15",
+    "@dedis/cothority": "^3.1.3",
     "@c4dt/dynacred": "dev",
     "@dedis/kyber": "^3.0.7",
     "core-js": "^2.6.9",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -24,7 +24,7 @@
     "@angular/platform-browser": "^7.2.15",
     "@angular/platform-browser-dynamic": "^7.2.15",
     "@angular/router": "^7.2.15",
-    "@dedis/cothority": "^3.1.3",
+    "@c4dt/cothority": "^3.1.15",
     "@c4dt/dynacred": "dev",
     "@dedis/kyber": "^3.0.7",
     "core-js": "^2.6.9",
--- a/webapp/src/app/admin/contacts/contacts.component.ts
+++ b/webapp/src/app/admin/contacts/contacts.component.ts
@@ -37,10 +37,10 @@ export class ContactsComponent implements OnInit {
     groups: DarcInstance[] = [];
 
     constructor(
-        private dialog: MatDialog,
-        private snackBar: MatSnackBar,
-        private bcvs: BcviewerService,
-        private location: Location,
+        protected dialog: MatDialog,
+        protected snackBar: MatSnackBar,
+        protected bcvs: BcviewerService,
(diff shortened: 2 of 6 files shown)


webapp/package.json
```diff
@@ -24,7 +24,7 @@
     "@angular/platform-browser": "^7.2.15",
     "@angular/platform-browser-dynamic": "^7.2.15",
     "@angular/router": "^7.2.15",
-    "@c4dt/cothority": "^3.1.15",
+    "@dedis/cothority": "^3.1.3",
```
tharvik: Wanted? Was introduced by #57 and `@dedis/cothority` is not NPM updated yet

webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -62,10 +62,22 @@ export class ContactsComponent implements OnInit {
...
-    async createContact(view?: string) {
-        const creds: IUserCred = {alias: "", email: "", view: "default", groups: []};
-        const groups = await this.uData.contact.getGroups();
-        creds.groups = groups.map((group) => group.darc.description.toString());
+    /**
+     * Pops up a dialog and asks for the alias and email of the new user to be created.
+     * If any of the view or groups parameter is the non-default, it will not be asked
+     * by the dialog.
+     *
+     * @param view - must be either "" or one of Data.views
```
tharvik: `view` has two meanings, either the name of the view to use (`!== ""`), or letting the user decide (`=== ""`). IMO, it would be best represented by the `string | undefined` type (and also forcing to unwrap it to be typesafe, as `Option` in functional languages)

webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -62,10 +62,22 @@ export class ContactsComponent implements OnInit {
...
+     * @param groups - must be either [] or one of the user's available groups
```
tharvik: So, either empty list or singleton list? Same argument as before then, using `string | undefined` looks better.

webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -37,10 +37,10 @@ export class ContactsComponent implements OnInit {
...
 
     constructor(
-        private dialog: MatDialog,
-        private snackBar: MatSnackBar,
-        private bcvs: BcviewerService,
-        private location: Location,
+        protected dialog: MatDialog,
+        protected snackBar: MatSnackBar,
+        protected bcvs: BcviewerService,
+        protected location: Location,
```
tharvik: All protected? I only see `this.dialog` used in PartnerComponent.

webapp/src/app/c4dt/partner/partner.component.ts
```diff
@@ -22,6 +24,15 @@ export class PartnerComponent extends ContactsComponent {
     }
```
tharvik: While I'm at it, I guess we can remove this block, as the super will already do that.

webapp/package.json
ineiti: Oups - I thought that was a leftover from me...

webapp/src/app/c4dt/partner/partner.component.ts
```diff
@@ -1,6 +1,8 @@
 import { Location } from "@angular/common";
 import { Component } from "@angular/core";
 import { MatDialog, MatSnackBar } from "@angular/material";
+import Log from "@dedis/cothority/log";
```
tharvik: Unused

PR #59, opened 2019-08-06
Nicer waiting window

ineiti:
Instead of using just a 'SnackBar' on the lower end, show a nice
modal, non-quittable window, with some animation of a blockchain
and new blocks being added.

Screenshot 2019-08-08 at 13 59 23


webapp/src/lib/Ui.ts
```diff
@@ -73,6 +102,121 @@ export interface IDialogConfig {
...
+    title: string;
+    worker: TWorker;
+}
+
+@Component({
+    selector: "app-dialog-transaction",
+    styleUrls: ["./app-dialog-transaction.scss"],
+    templateUrl: "dialog-transaction.html",
+})
+export class DialogTransactionComponent implements OnInit {
```
tharvik: Hum, `UI.ts` is going over 200 lines, maybe extracting a/some/this classes to other files?

webapp/src/lib/Ui.ts
```diff
@@ -73,6 +102,121 @@ export interface IDialogConfig {
...
+    }
+
+    percentage: number;
+    blockIndex: number;
+    blocks: Element[] = [];
+    transaction: Element;
+    error: string;
+    text: string;
+    @ViewChild("main", {static: false}) main?: ElementRef;
+    ubTimer: Timer;
```
tharvik: Why having it public? Going private will avoid cross-contamination/misuse.

webapp/src/lib/Ui.ts
```diff
@@ -73,6 +102,121 @@ export interface IDialogConfig {
...
+        const prog = (p: number, t: string) => DialogTransactionComponent.progress(this, p, t);
+        try {
+            await this.data.worker(prog);
+            prog(-100, "Done");
+        } catch (e) {
+            this.error = e.toString();
+        }
+    }
+
+    addBlock(index: number = this.blockIndex): Element {
```
tharvik: Hum, as you're doing it with a dedicated function, I think that you already tried with a template and `ngIf` magics; but, if not, that might be doable/clearer with it.

webapp/src/lib/Ui.ts
```diff
@@ -73,6 +102,121 @@ export interface IDialogConfig {
...
+    blocks: Element[] = [];
+    transaction: Element;
+    error: string;
+    text: string;
+    @ViewChild("main", {static: false}) main?: ElementRef;
+    ubTimer: Timer;
+
+    constructor(
+        public dialogRef: MatDialogRef<DialogTransactionComponent>,
+        public bcv: BcviewerService,
```
tharvik: And some `readonly` are pretty useful also, to avoid modifying fields not meant to.

webapp/src/lib/dialog-transaction.html
```diff
@@ -0,0 +1,20 @@
...
+        <img src="assets/imgs/cothority.png" class="cothority"/>
+
+        <div class="progress-box">
+            <div [style.width.%]="percentage" class="progress-bar">
+                {{text}}
+            </div>
+        </div>
+    </div>
+    <div *ngIf="error">
+        <div class="error" >Error: {{ error }}</div>
```
tharvik: So yeah, let's choose the syntax we want for name reference in templates.
From what I read, the `{{ var }}` is used in the reference and emphasize the visibility of the reference. `{{var}}` is also used, as a more concise form.
Mixing both make it a bit harder to read. WDYT?

webapp/src/lib/Ui.ts
```diff
@@ -73,6 +102,121 @@ export interface IDialogConfig {
...
+}
+
+@Component({
+    selector: "app-dialog-transaction",
+    styleUrls: ["./app-dialog-transaction.scss"],
+    templateUrl: "dialog-transaction.html",
+})
+export class DialogTransactionComponent implements OnInit {
+
+    static progress(dtc: DialogTransactionComponent, percentage: number, text: string) {
```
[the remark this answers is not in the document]
ineiti: as the `progress` is passed to the calling method, `this` changes to the calling method... Stupid, but true. In fact it started out non-static, but then it failed. So I made this 'this-catcher'. Some people use `self` and cast `this` to it, to be sure to have the correct `this`...

webapp/src/lib/Ui.ts
ineiti: That would be more of a `ngIf`, `ngClass` orgy. Possible, but I thought it's simpler to create the objects and add the classes here.

webapp/src/lib/Ui.ts
```diff
@@ -47,6 +49,33 @@ export async function showDialogInfo(dialog: MatDialog, title: string, text: str
...
+    const tc = dialog.open(DialogTransactionComponent, {
+        data: {
+            title,
+            worker,
+        },
+        disableClose: true,
+    });
+
+    return new Promise((resolve) => {
+        tc.afterClosed().subscribe((res) => {
```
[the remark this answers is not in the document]
ineiti: Yes, but as `reject` is not used here, I only use `resolve`.

webapp/src/lib/Ui.ts
```diff
@@ -47,6 +49,33 @@ export async function showDialogInfo(dialog: MatDialog, title: string, text: str
...
+    return showTransactions(dialog, title, async (progress: TProgress) => {
+        progress(50, "Storing Credential");
+        await store();
+    });
+}
+
+export type TProgress = (percentage: number, text: string) => void;
+export type TWorker = (progress: TProgress) => Promise<void>;
+
+export async function showTransactions(dialog: MatDialog, title: string, worker: TWorker): Promise<string> {
```
tharvik: Why is it returning the result of the dialog? There is not input by the user, correct? Returning the potentially long-to-wait result would be more useful IMO.

webapp/src/lib/dialog-transaction.html
ineiti: Oh my - I'm always using <CMD><OPTION>L in IntelliJ and just go with whatever he proposes to me...

Let's go with `{{ error }}` - and we should find a formatter for the html, too...

webapp/src/lib/Ui.ts
ineiti: As the dialog catches the error of the callback, I return the error here, if there is one. This looked better, as the dialog itself already shows the error to the user, so I didn't want to `throw` it again.

tharvik: Ho, right, for the error case, we don't need to throw it also, but for the happy path, we might want to retrieve the result.
To better understand why I'm thinking of using it this way: in `admin/contacts/contacts.components.ts:ContactsComponent.darcInstanceAdd`, instead of wrapping it all in callback style, awaiting on some result of the callback from `storeCredential` allows to separate the logic of "letting the user choose a darc instance from a list" from the "where to store the actions".

webapp/src/lib/Ui.ts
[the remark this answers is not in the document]
tharvik: Hum, I've tried to move it to a member function and it seems to work. I think by binding `this` param in the lambda in `startTransactions.prog`, it avoids the issue you mentionned.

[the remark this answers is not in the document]
ineiti: I wonder why it didn't work before and what I changed in the meantime...

webapp/src/lib/Ui.ts
```diff
@@ -24,42 +32,79 @@ export async function showSnack(snack: MatSnackBar, text: string, cmd: () => voi
...
+/**
+ * Convenience method for showTransactions that only shows one progress step: "Storing Credential".
+ *
+ * @param dialog reference to the matDialog
+ * @param title shown in h1 in the dialog
+ * @param store the callback to the actual storing of the credential.
+ */
 export async function storeCredential(dialog: MatDialog, title: string,
-                                      store: () => Promise<void>) {
+                                      store: <T>() => Promise<T | void>) {
```
tharvik: `T | void` is weird to read (and should typecast to `T` anyway). As you're `return`ing the result of `store` also,  `storeCredential` should become `storeCredential<T>` (the type is not bound to `store` only in this case).

```suggestion
                                      store: () => Promise<T>): Promise<T> {
```

webapp/src/lib/Ui.ts
```diff
@@ -24,42 +32,79 @@ export async function showSnack(snack: MatSnackBar, text: string, cmd: () => voi
...
-        await store();
+        return store();
     });
 }
 
+// Progress type to be used in showTransactions.
 export type TProgress = (percentage: number, text: string) => void;
-export type TWorker = (progress: TProgress) => Promise<void>;
+// Worker callback that implements multiple steps and calls progress before each step.
+export type TWorker<T> = (progress: TProgress) => Promise<T | void>;
```
tharvik:
```suggestion
export type TWorker<T> = (progress: TProgress) => Promise<T>;
```

webapp/src/lib/Ui.ts
```diff
@@ -70,11 +115,17 @@ export async function showTransactions(dialog: MatDialog, title: string, worker:
 
     return new Promise((resolve) => {
         tc.afterClosed().subscribe((res) => {
-            resolve(res);
+            resolve(res as T);
```
tharvik: Casting, especially in functional world, is indicative of a code smell. Why would you need that (it works if I remove it)?

webapp/src/lib/dialog-transaction.ts
```diff
@@ -72,16 +52,25 @@ export class DialogTransactionComponent implements OnInit {
...
+                this.dialogRef.close(result);
+            }, 1000);
         } catch (e) {
             this.error = e.toString();
         }
     }
 
     addBlock(index: number = this.blockIndex): Element {
+        if (!this.main) {
+            return null;
```
tharvik: Hum, so, `null` and `undefined`, I don't know why there is two bottom types and when to use each; from the typescript's [coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined), they prefers to use `undefined` so this is what I use. Do you have any preference/idea/limitation on each?

webapp/src/lib/Ui.ts
ineiti: `storeCredential` is a convenience function that only does one thing: storing the credential in the blockchain. So I replaced the callback with a parameter of type `Data`. While replacing all calls to `storeCredential`, I couldn't resist to refactor `ContactsComponent`, too...

webapp/src/lib/Ui.ts
ineiti: Most of the calls to `showTransaction` don't return anything, this is why I would like to keep the `void`...

webapp/src/lib/Ui.ts
ineiti: oups - leftover of me getting acquainted with generics...

webapp/src/lib/dialog-transaction.ts
ineiti: Oh my - in your hands even typescript is safe!

webapp/src/lib/Ui.ts
tharvik: Yup, and `T` already ~contains `void`, as it represent any type (there is no type bounds). The compiler reduce `T | void` to `T` anyway.

ineiti: yes, but then I need to

```
return void/null/undefined
```

at the end of my callbacks :(

ineiti: You're always right, dear...

PR #60, opened 2019-08-07
Corrected navigation

ineiti:
When signing up or creating new device, the navigation should
go to the landing page.

Also corrected the authorization of OCS when testing locally.


dynacred/src/Data.ts
```diff
@@ -158,7 +158,7 @@ export class Data {
             signers: [adminSigner],
         });
 
-        const ocs = new OnChainSecretRPC(bc);
+        await (new OnChainSecretRPC(bc)).authorizeRoster();
```
tharvik: Got "1002 unexpected error: ByzCoinID already authorised" when using this path twice

webapp/package.json
```diff
@@ -24,9 +24,9 @@
     "@angular/platform-browser": "^7.2.15",
     "@angular/platform-browser-dynamic": "^7.2.15",
     "@angular/router": "^7.2.15",
-    "@c4dt/cothority": "^3.1.15",
     "@c4dt/dynacred": "dev",
-    "@dedis/kyber": "^3.0.7",
+    "@dedis/cothority": "^3.1.4",
```
tharvik: Wanted I guess now, but just checking

ineiti: Yes, this is the latest npm as of today - does it now have all the changes from the c4dt one?

tharvik: It has the one I wanted, was `getProofFromLatest` used in c4dt and not merged in dedis?

ineiti: in the end the `getProofFromLatest` was not applicable in the `dedis/cothority` code, as it was all proofs that needed to come from the genesis anyway, because it's LTS/Calypso, which needs the full proof.

So I removed that PR: https://github.com/dedis/cothority/pull/2031

PR #65, opened 2019-08-20
Doing more checks

ineiti:
Adding some checks for the conode-docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,23 +21,26 @@ jobs:
         - make docker
         - make docker_run &
         - make conode
+        - make docker_wait
         - build/bcadmin latest --bc bc*cfg
-        - docker rm -f conode
     - name: dynacred
       language: node_js
-      node_js: 10
+      node_js: "10"
       install: cd dynacred && npm ci
       script: npm run test
     - name: webapp
       language: node_js
-      node_js: 10
+      node_js: "10"
       install:
+        - gimme 1.12.4
+        - . $HOME/.gimme/envs/go1.12.4.env
         - cd webapp && npm ci
         - npm link ../dynacred
       before_script: npm run lint
       script:
         - npm run build --prod
-        - docker run --rm -d -p 7770-7777:7770-7777 --name conode c4dt/conode:dev
+        - make -C ../conode docker_run &
+        - make -C ../conode docker_wait
         - npm run test --no-watch
 
     - stage: deploy
@@ -50,7 +53,7 @@ jobs:
         script: ssh omniledger@c4dtsrv1.epfl.ch bin/update.sh
     - name: NPM
       language: node_js
-      node_js: 10
+      node_js: "10"
       script: skip # node_js.script default to `make test`
       before_deploy: echo "//registry.npmjs.org/:_authToken=${DEPLOY_NPM_TOKEN}" > "$HOME/.npmrc"
       deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ jobs:
       script:
         - cd conode
         - make docker
-        - docker run --rm -t -p 7770-7777:7770-7777 --name conode c4dt/conode:dev &
+        - make docker_run &
         - make conode
         - build/bcadmin latest --bc bc*cfg
         - docker rm -f conode
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ jobs:
       script:
         - cd conode
         - make docker
+        - docker run --rm -t -p 7770-7777:7770-7777 --name conode c4dt/conode:dev &
+        - make conode
+        - build/bcadmin latest --bc bc*cfg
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -19,6 +19,10 @@ jobs:
       script:
         - cd conode
         - make docker
+        - docker run --rm -t -p 7770-7777:7770-7777 --name conode c4dt/conode:dev &
+        - make conode
+        - build/bcadmin latest --bc bc*cfg
```
tharvik: Concurrency issue, `docker run` may not have fully started yet. As ws is the main mean of communication and answers to HTTP, I used some curl loop in [c4dt/services](https://github.com/c4dt/services/blob/master/mk/service/backend.mk#L80).

.travis.yml
```diff
@@ -19,6 +19,10 @@ jobs:
       script:
         - cd conode
         - make docker
+        - docker run --rm -t -p 7770-7777:7770-7777 --name conode c4dt/conode:dev &
```
tharvik:
```suggestion
        - make docker_run &
```

.travis.yml
```diff
@@ -19,6 +19,10 @@ jobs:
...
+        - make conode
+        - build/bcadmin latest --bc bc*cfg
+        - docker rm -f conode
```
tharvik: Why? It's the end of this run

.travis.yml
```diff
@@ -33,6 +37,7 @@ jobs:
       before_script: npm run lint
       script:
         - npm run build --prod
+        - docker run --rm -d -p 7770-7777:7770-7777 --name conode c4dt/conode:dev
```
tharvik:
```suggestion
        - make docker_run &
```

conode/Makefile
```diff
@@ -11,7 +11,8 @@ cothority:
 conode: cothority
 	@( cd cothority; git pull )
 	echo "Compiling binaries"; \
-	GO111MODULE=on go build -o build/conode; \
+	export GO111MODULE=on; \
```
tharvik: So, I saw a lot of `GO111MODULE` defined around; from the [golang wiki](https://github.com/golang/go/wiki/Modules#how-to-install-and-activate-module-support), it seems that it's not needed in our cases.
And so, I've encountered the same issue as you, if one of its deps is not a module, it fails (for example with https://github.com/qantik/qrgo); maybe fixing upstream (it's from someone who was at DEDIS)?

.travis.yml
ineiti: I think this will not work, as this is in the `webapp` directory, and this one has no `docker_run` target.

.travis.yml
ineiti: No cleaning up?

.travis.yml
tharvik: Ho, you're right, I was a bit quick
```suggestion
        - make -C ../conode docker_run &
```

conode/Makefile
ineiti: True - IIRC, if the .go file is not in the `$GOPATH/src/$(head -n 1 go.mod)` the module should be activated anyway.

But usually the qantik-repo should be included correctly in the `go.mod` file, so I don't really understand why it fails here?

.travis.yml
ineiti: Oh my - this should still be ported completely to your services...

tharvik: Yup; hopefully, it will help to better know/understand what's happening. Maybe after finishing the CAS login?

.travis.yml
tharvik: Yup, no need to clean, as the server is taken down by travis anyway

conode/Makefile
tharvik: > But usually the qantik-repo should be included correctly in the go.mod file, so I don't really understand why it fails here?

I don't know either, maybe a missing `go.mod` in the repo? That's the only guess I have, you're the go master here :)

.travis.yml
ineiti: Added a `docker_wait` target to `conode/Makefile`

.travis.yml
```diff
@@ -19,20 +19,32 @@ jobs:
...
       language: node_js
-      node_js: 10
+      node_js: "10"
       install:
+        - gimme 1.12
+        - go version
+        - gimme 1.12.4
+        - go version
+        - . $HOME/.gimme/envs/go1.12.4.env
+        - go version
```
tharvik: Debug print?

.travis.yml
```diff
@@ -19,20 +19,32 @@ jobs:
...
       install: cd dynacred && npm ci
       script: npm run test
     - name: webapp
       language: node_js
-      node_js: 10
+      node_js: "10"
       install:
+        - gimme 1.12
+        - go version
+        - gimme 1.12.4
```
tharvik: Why choosing two version?

PR #71, opened 2019-08-27
Show QRCode for new devices

ineiti:
In order to copy a device to the mobile app, it shows now the QRCode of the URL.
Additionally, the 'Secret' tab got replaced by a 'Personhood' tab that will show
some information about the personhood-data.

Deps on #65

### DIFF
--- a/webapp/src/app/app.component.ts
+++ b/webapp/src/app/app.component.ts
@@ -4,7 +4,6 @@ import { Router } from "@angular/router";
 
 import Log from "@c4dt/cothority/log";
 
-import { Data } from "@c4dt/dynacred";
 import { BcviewerService } from "./bcviewer/bcviewer.component";
 import { UserData } from "./user-data.service";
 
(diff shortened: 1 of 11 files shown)


webapp/src/app/app.component.ts
```diff
@@ -50,12 +51,21 @@ export class AppComponent implements OnInit {
...
                         this.loading = false;
-                        await this.router.navigate(["/newuser"]);
+                        return this.newUser();
                     }
                 });
             }
         }
     }
+
+    async newUser() {
```
tharvik: `private`, what's the return type (it's boolean but it is not used/relevant so maybe tagging it at `Promise<void>`)?

webapp/src/app/app.component.ts
```diff
@@ -4,6 +4,7 @@ import { Router } from "@angular/router";
 
 import Log from "@c4dt/cothority/log";
 
+import { Data } from "@c4dt/dynacred";
```
tharvik: I don't think you need it, you access it with `uData`.

PR #74, opened 2019-08-29
Using url-parse also in cothority-ts

ineiti:


mobile/app/lib/cothority/network/proto.ts
```diff
@@ -172,21 +173,22 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
                 break;
             }
             case "https:": {
-                urlParser.protocol = "wss:";
+                proto = "wss:";
                 break;
             }
             default : {
                 throw new Error("The url field should use either 'http:' or 'https:', but we found "
                     + urlParser.protocol);
```
tharvik:
```suggestion
                     + proto);
```

mobile/app/lib/cothority/network/proto.ts
```diff
@@ -172,21 +173,22 @@ export class ServerIdentity extends Message<ServerIdentity> {
...
             }
         }
-        let result = urlParser.toString();
+        let result = url.replace(/^https?:/, proto);
```
tharvik: If you modify `urlParser` directly, no need to string replace stuff.

mobile/app/lib/cothority/network/proto.ts
ineiti: why? This is the error-message, so `proto` is not set.

mobile/app/lib/cothority/network/proto.ts
ineiti: That was the original `new URL`, but the `url-parser` doesn't allow to set the `protocol`, this is why I changed it here to use the `proto` variable.

mobile/app/lib/cothority/network/proto.ts
tharvik: Oups, you're correct

mobile/app/lib/cothority/network/proto.ts
tharvik: Hum, it seems that there is a [URL.set](https://www.npmjs.com/package/url-parse#urlsetkey-value) to do that, maybe something like that?
```suggestion
        urlParser.set("protocol", proto);
```

ineiti:
Screw `url-parse`, this is a job for a regexp...

PR #75, opened 2019-08-29
revamp cas login

tharvik:
Reimplement CAS support.

* `serve_all.py` is gone, we now use a gin powered server
* tickets are generated as a coin exchanged back and forth between the user and the `Action` to login with
* tickets are validated by looking from the most recent transaction until it's found
* `Action` creation also create a `CoinInstance`
* add small idiomatic test system
---
TODO
- [x] limit back search of blocks
- [x] fully check tx in block
- [x] return the wanted username (how to choose?)


webapp/cas/go.mod
```diff
@@ -0,0 +1,12 @@
+module gopkg.in/c4dt/omniledger.v1/webapp/cas
```
ineiti: don't mix gopkg.in and go.mod - they're two different solutions to the same problem. It's `github.com/dedis/comniledger/webapp/cas`

webapp/cas/config.go
```diff
@@ -0,0 +1,141 @@
...
+		serviceToCoinInstanceIDs[k], err = hex.DecodeString(v)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	txValidityDuration, err := time.ParseDuration(tomlConf.TxValidityDuration)
+
+	return &Config{
+		byzCoinID,
```
ineiti: I think `golint` will complain here that you don't name the fields. And if it doesn't complain, I will ;) I got used to it, because, even if most of the variables correspond to the field names, not all do. So it's nice to have the field name with it.

webapp/cas/login.uml
```diff
@@ -0,0 +1,29 @@
+@startuml
```
ineiti: Can you write a comment in the file how to view this? If it converts to a .png, add it to a Makefile that creates the png?

webapp/cas/main.go
```diff
@@ -0,0 +1,85 @@
+package main
+
+import (
+	"io/ioutil"
```
ineiti: Please don't use tabs as indention, but 2 spaces. I'm surprised `gofmt` lets this pass...

webapp/cas/main.go
```diff
@@ -0,0 +1,85 @@
...
+	r := gin.Default()
+
+	r.GET("/api/v0/cas/p3/serviceValidate", func(c *gin.Context) {
+		c.XML(http.StatusOK, cas.ServiceValidateXML(
+			c.Query("service"), c.Query("ticket"),
+		))
+	})
+
+	r.GET("/", redirectToRoot)
+	r.GET("/register/*path", redirectToRoot)
```
ineiti: That feels awkward - if we add a new path in the webapp, we need to add it also in the go? Can't you have a 'catch-all', and then prioritise?

webapp/cas/main.go
```diff
@@ -0,0 +1,85 @@
...
+		log.Fatal(err)
+	}
+
+	conf, err := ParseConfig(toml)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	return NewCAS(*conf)
+
```
ineiti: empty line

webapp/cas/service_validate.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (cas CAS) containsLoginProof(coinInstID skipchain.SkipBlockID, block skipchain.SkipBlock, hashedChallenge []byte) (bool, error) {
+	logI := func(e ...interface{}) { log.Print(e...) }
+	logE := func(e ...interface{}) {
+		buf := make([]interface{}, 1+len(e))
+		buf[0] = "!! "
+		copy(buf[1:], e)
+		log.Print(buf...)
+	}
+	logI("checking block: ", block.Hash.Short())
```
ineiti: Please use the `onet.log` calls.

webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -264,10 +264,43 @@ export class ContactsComponent implements OnInit {
...
+                ...this.uData.spawnerInstance.spawnDarcInstructions(
+                    this.uData.coinInstance,
+                    signDarc, coinDarc),
+                ...this.uData.spawnerInstance.spawnCoinInstructions(
+                    this.uData.coinInstance,
+                    coinDarc.id,
+                    coinID));
+            await transaction.updateCountersAndSign(this.uData.bc, [[this.uData.keyIdentitySigner]]);
+
+            progress(60, "Creating new DARCs");
```
ineiti:
```suggestion
            progress(60, "Creating new DARCs and coin");
```

webapp/src/app/api/v0/cas/login/login.component.ts
```diff
@@ -1,55 +1,133 @@
...
     }
 
     async deny() {
         await this.router.navigate(["/"]);
     }
 
+    /**
+     * Try to put the challenge on the chain
+     *
+     * It's encoded as a coin echange back and forth from the user to the Action's CoinInstance.
```
ineiti:
```suggestion
     * It's encoded as a coin exchange back and forth from the user to the Action's CoinInstance.
```

webapp/src/app/api/v0/cas/login/login.component.ts
```diff
@@ -1,55 +1,133 @@
...
+     * The Instruction getting it back contains an Argument with the hash of challenge, so that when the user provides
+     * it, it proves that he was the one generating it.
```
ineiti:
```suggestion
     * it, the transaction proves that he was the one generating it.
```

ineiti:
I didn't find any test infrastructure for the cas login - did I miss something?

webapp/cas/client.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"go.dedis.ch/cothority/v3/byzcoin"
+	"go.dedis.ch/cothority/v3/skipchain"
+	"go.dedis.ch/onet/v3"
+)
+
+type clients struct {
```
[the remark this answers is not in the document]
tharvik: How do I extract the `skipchain.Client` from the `byzcoin.Client` (it contains a `onet.Client` only)?

webapp/cas/main.go
tharvik: From the [Effective go documentation](https://golang.org/doc/effective_go.html#formatting)
> Indentation
We use tabs for indentation and gofmt emits them by default. Use spaces only if you must.

webapp/cas/main.go
tharvik: Sadly, no, as gin is *very* fast, it uses a single Trie without fallback. I might be able to work it around by adding a preprocessor to every request, if you need

webapp/cas/go.mod
tharvik: *c4dt ;)

webapp/cas/config.go
tharvik: (it's not complaining :p) It's not the same meaning, using named composite literals set the zero values for missing keys, ie doesn't msg/warn/fail when adding new fields (silent fail IMHO).

From the [Effective go documentation](https://golang.org/doc/effective_go.html#composite_literals)
> The fields of a composite literal are laid out in order and must all be present. However, by labeling the elements explicitly as field:value pairs, the initializers can appear in any order, with the missing ones left as their respective zero values.

tharvik:
> If I understood correctly, there is currently no way to update existing actions to contain a coin, is that correct?

You're correct, only new action generates a coin instance.

tharvik:
Testing added, local test done.

PR #96, opened 2019-09-10
I18n improvements

cgrigis:
* More strings internationalized in the XML templates, with English and French locales.
* Updates to the existing French translation.


cgrigis:
**CAUTION**: I have not yet tested the changes, as my machine is not yet setup for a build (must install Android SDK).

PR #116, opened 2019-09-19
Correctly wait for first block

ineiti:
When a new transaction is shown, DialogTransactionComponent needs
to make sure that there is already a block in the cache, else
it needs to wait.

### DIFF
--- a/webapp/src/app/bcviewer/bcviewer.component.ts
+++ b/webapp/src/app/bcviewer/bcviewer.component.ts
@@ -21,6 +21,7 @@ import { UserData } from "../user-data.service";
 export class BcviewerService {
 
     @Output() update: EventEmitter<boolean> = new EventEmitter();
+    @Output() newStatus: EventEmitter<boolean> = new EventEmitter();
     currentBlock: SkipBlock;
 
     updateBlocks() {
@@ -68,6 +69,7 @@ export class BcviewerComponent implements OnInit {
                 this.blocks.splice(0, this.blocks.length - 4);
             }
             this.showBlockService.currentBlock = this.blocks[this.blocks.length - 1].sb;
+            this.showBlockService.newStatus.emit(true);
         }
     }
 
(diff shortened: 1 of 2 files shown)


webapp/src/app/bcviewer/bcviewer.component.ts
```diff
@@ -21,6 +21,7 @@ import { UserData } from "../user-data.service";
 export class BcviewerService {
 
     @Output() update: EventEmitter<boolean> = new EventEmitter();
+    @Output() newStatus: EventEmitter<boolean> = new EventEmitter();
```
tharvik: There is only one possible value (`true` in this case), so this doesn't require a two values type, better use `void` (which contains none but the importance is that the event is triggered).
```suggestion
    @Output() newStatus: EventEmitter<void> = new EventEmitter();
```

webapp/src/app/bcviewer/bcviewer.component.ts
```diff
@@ -68,6 +69,7 @@ export class BcviewerComponent implements OnInit {
                 this.blocks.splice(0, this.blocks.length - 4);
             }
             this.showBlockService.currentBlock = this.blocks[this.blocks.length - 1].sb;
+            this.showBlockService.newStatus.emit(true);
```
tharvik: With void
```suggestion
            this.showBlockService.newStatus.emit();
```

webapp/src/lib/dialog-transaction.ts
```diff
@@ -34,12 +34,21 @@ export class DialogTransactionComponent<T> implements OnInit {
     }
 
     async ngOnInit() {
-        this.bcv.updateBlocks();
+        this.updateBlocks();
+        if (!this.bcv.currentBlock) {
+            await new Promise((resolve) => {
+                this.bcv.newStatus.subscribe((update) => {
+                    if (update) {
```
tharvik: With void, no need to check for `update` as firing is enough to know that.
This promise block can be reduced to `await new Promise(this.bcv.newStatus.subscribe);` (if I'm correct).

ineiti: That gives:
```
ERROR Error: "Uncaught (in promise): TypeError: this is undefined
```
I also tried 
```
await this.bcv.newStatus.toPromise();
```
which doesn't work, neither.

tharvik: Haa sure, `this` issue; nevermind then.

PR #134, opened 2019-09-27
webapp/login/cas: improve style

tharvik:
Show transaction when pushing challenge, center the `/login`.


ineiti:
That was easy ;)

PR #141, opened 2019-09-30
webapp/UI: fix showTx handling of Promise

tharvik:
Didn't find a way to gen observable error inside a DialogRef, so any error return is really an error.


ineiti:
Failing tests seem to be wrong tests that use the spawnerInstance to create darcs with `spawn:calypsoRead` as a rule, which is not allowed anymore...

PR #144, opened 2019-10-01
webapp/cas/stress: use headless chrome, small fixes

tharvik:
* headless is better
* go is weird with scoping and closures

### DIFF
--- a/webapp/cas/cmd/cas-stress/main.go
+++ b/webapp/cas/cmd/cas-stress/main.go
@@ -34,18 +34,16 @@ func main() {
 			return err
 		}
 
-		driver := agouti.ChromeDriver()
+		driver := agouti.ChromeDriver(agouti.ChromeOptions("args", []string{"headless"}))
 		if err := driver.Start(); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


webapp/cas/cmd/cas-stress/main.go
```diff
@@ -34,18 +34,20 @@ func main() {
...
-		driver := agouti.ChromeDriver()
+		driver := agouti.ChromeDriver(agouti.ChromeOptions("args", []string{"headless"}))
 		if err := driver.Start(); err != nil {
 			return err
 		}
 
 		clickers := make([]func(*agouti.Page) error, 0)
 		url := c.String("with-matrix-url")
 		if url != "" {
+			url := url
```
ineiti: I think `golint` (or tslint) would choke on this as you're shadowing an existing variable.

Why do you need this in the first place?

You could do something like

```
if url := c.String("..."); url != ""{
}
```

tharvik: Ho, right, I have a bad habit of forgetting this go construct

PR #146, opened 2019-10-02
dynacred/Contact: std prepareInitialCred, rm unused get/set

tharvik:
* rather use `undefined` than `null`
* remove some useless constructs
* now works with https://github.com/dedis/cothority/pull/2041

### DIFF
--- a/dynacred/src/Badge.ts
+++ b/dynacred/src/Badge.ts
@@ -41,8 +41,7 @@ export class Badge {
             await this.party.partyInstance.mine(d.keyPersonhood._private.scalar,
                 null, darc);
             // Setting spawner-id
-            d.contact.credential = Contact.prepareInitialCred(d.alias, d.keyIdentity._public, d.spawnerInstance.id,
-                null, null);
+            d.contact.credential = Contact.prepareInitialCred(d.alias, d.keyIdentity._public, d.spawnerInstance.id);
             d.spawnerInstance = await SpawnerInstance.fromByzcoin(d.bc, d.spawnerInstance.id);
             // Use the coin and the darc to create a new user
             const ci = await CoinInstance.fromByzcoin(d.bc, CoinInstance.coinIID(darc.getBaseID()));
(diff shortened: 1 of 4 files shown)


ineiti:
I think you're removing stuff that should _not_ be removed. Also, removing unused get/set is not cool - as you said yourself, sometimes it's used in an app that depends on `dynacred`, so just removing it will fail.

Also removing the versioning of the credential-struct is - what shall I say - uncool. Why? It is used, and will be used (when moving to the phonebook).

Please let's get together and look at this from a bigger point of view.

ineiti:
Moving it to TODO until we discussed it.

tharvik:
> * don't remove 'unused' getters/setters, you don't know all the apps.

So, waiting for mobile to be merged? We might have to discuss about API stability at some point.

> * don't remove the versioning of the credential-structure - if you have a better solution, let's discuss it, but just removing it will not work and give us problems if we want to move forward

It is unused currently, from what I saw in code

 * version 0 is initial setup via `prepareInitialCred`
 * version 1 is done when updating, adding a `("1-devices", "initial")` (but it was already added in version 0)

So, only version 0 is relevant and the simple existence of the structure can be seen as version 0.

> Moving it to TODO until we discussed it.

Sure, that'll be easier face-to-face :)

tharvik:
At least https://github.com/c4dt/omniledger/pull/146/commits/e0a994f2a935d16efe3f6066600c84be89c9ee35 needs to go in, as using `null` to denote missing values is wrong since https://github.com/dedis/cothority/pull/2041

ineiti:
> At least [e0a994f](https://github.com/c4dt/omniledger/commit/e0a994f2a935d16efe3f6066600c84be89c9ee35) needs to go in, as using `null` to denote missing values is wrong since [dedis/cothority#2041](https://github.com/dedis/cothority/pull/2041)

can you make a PR with that commit?

tharvik:
Outdated.

PR #151, opened 2019-10-09
webapp/cas: bump deps

tharvik:
* go.dedis.ch/cothority v3.3.0 is released, we can refollow master


ineiti:
Question from a noob: can `go.sum` be cleaned up? Should it?

tharvik:
Hum, from what I read, it's like the `package-lock.json`, it should be commited when changing the `go.mod`.

ineiti:
My question was more like: why doesn't `go mod -sync` exist anymore in go 1.13?

At least `package-lock.json` doesn't keep track of _every_ version that has been _ever_ used in the project...

tharvik:
> My question was more like: why doesn't `go mod -sync` exist anymore in go 1.13?

```
tharvik@localhost cas % go mod -sync
go: go mod -sync is now go mod tidy
```

ineiti:
I think we should add that to the travis `build`-target. `go mod tidy` and then check that `go.mod` and `go.sum` are still the same...

tharvik:
https://github.com/c4dt/omniledger/issues/152

PR #153, opened 2019-10-11
i18n for add_party dialog

cgrigis:
A leftover i18n that I still had in my tree. :)


ineiti:
Closed in favor of #221

PR #156, opened 2019-10-24
webapp/api/login: deploy on c4dt.org

tharvik:
* rm old WordPress login
* use correct URL for requests coming from c4dt's WordPress


tharvik:
> Stupid question: will that also work for someone going through `https://c4dt.org`, without the `www`?

Nope, but there is not way to load WordPress' content using solely `c4dt.org`; the chain of redirect is http://c4dt.org -> https://c4dt.org -> https://www.c4dt.org

PR #158, opened 2019-10-24
Ios fix

ineiti:
- Some cosmetic changes
- updated nativescript-websockets so it works on iOS13


.travis.yml
```diff
@@ -16,6 +16,8 @@ jobs:
       name: conode
       language: go
       go: "1.12.x"
+      install:
+        - export GO111MODULE=on
```
tharvik: It's already ~fixed by #157.

mobile/app/App_Resources/Android/src/main/res/values/strings.xml
```diff
@@ -153,9 +154,9 @@
   <string name="_score_header_attributes_1IpA9K">"Attributs (%1$d%%)"</string>
   <string name="_score_done_1O5zTM">"Tu as complété les %1$d %2$s. %3$s"</string>
   <string name="_score_improve_mDxz3">"Tu as fait %1$d sur %2$d %3$s. %4$s"</string>
-  <string name="_score_desc_1O5xmK">"Ton degré d\'individualité est une évaluation faite par le système que tu es bien un individu unique. Plus ce degré est élevé, plus le système sera confiant que tu ne va pas l\'attaquer en te faisant passer pour plusieurs personnes."</string>
-  <string name="_score_header_W8cDM">"Degré d\'individualité"</string>
-  <string name="_score_title_wH00B">"Individualité"</string>
+  <string name="_score_desc_1O5xmK">"L\'accomplissement montre combien des functionlaités de l\'app tu as déjà utilisé. Plus tu utilises les différentes functionalités, plus ton accomplissement sera élevé."</string>
```
tharvik:
```suggestion
  <string name="_score_desc_1O5xmK">"L\'accomplissement montre combien des functionnalités de l\'app tu as déjà utilisé. Plus tu utilises les différentes functionalités, plus ton accomplissement sera élevé."</string>
```

mobile/app/lib/nativescript-ws.ts
```diff
@@ -29,7 +29,7 @@ export class NativescriptWebSocketAdapter extends WebSocketAdapter {
 
     /** @inheritdoc */
     onMessage(callback: (data: Buffer) => void): void {
-        this.ws.on("message", (socket, msg) => {
+        this.ws.on("message", (socket, msg: ArrayBuffer) => {
             if (msg instanceof Buffer || msg instanceof ArrayBuffer) {
```
tharvik: as `msg` is an ArrayBuffer, it will always be true.

ineiti: I'm not sure that `nativescript-websockets` always gives an `ArrayBuffer` here - I'll remove the method signature and keep the test...

mobile/app/App_Resources/Android/src/main/res/values/strings.xml
ineiti: Verify the verifier: Il y a encore un 'o' dans fonctionnalité...

.travis.yml
ineiti: I will not use cothority/v4 in this branch. Once this is in, my next step is to merge it with `master`, then it'll use `v4`...

mobile/app/App_Resources/Android/src/main/res/values/strings.xml
tharvik: Good catch; edited

PR #159, opened 2019-10-28
{conode,webapp/cas}: revert to v3

tharvik:
Reverting to cothority v3


ineiti:
Sorry for that back-and-forth...

PR #162, opened 2019-11-27
Correctly display newuser page

### LINKED ISSUE #160: New User landing error
When a new user without an ID goes to demo.c4dt.org/omniledger, there is an error-message. It should go straight to `/newuser`.

ineiti:
Closes #160

### DIFF
--- a/dynacred/src/Storage.ts
+++ b/dynacred/src/Storage.ts
@@ -21,7 +21,7 @@ export class StorageDB {
 
     static async get(key: string): Promise<string> {
         const entry = await this.getDB().get(key);
-        if (entry) {
+        if (entry !== undefined) {
             return entry.buffer;
         }
         return "";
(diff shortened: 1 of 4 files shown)


dynacred/src/Storage.ts
```diff
@@ -32,13 +32,22 @@ export class StorageDB {
     }
 
     static async getObject(entry: string): Promise<any> {
```
tharvik: A common test of the return value is `Object.keys(got).length > 0`, which is to check for the absence of relevant value. Using `undefined` explicit the missing object.
 
```suggestion
     static async getObject(entry: string): Promise<any | undefined> {
```

dynacred/src/Storage.ts
```diff
@@ -21,7 +21,7 @@ export class StorageDB {
 
     static async get(key: string): Promise<string> {
```
tharvik: See `getObject` comment.

```suggestion
     static async get(key: string): Promise<string | undefined> {
```

dynacred/src/Storage.ts
```diff
@@ -32,13 +32,22 @@ export class StorageDB {
...
+            return {};
+        }
+        try {
+            const obj = JSON.parse(d, (key, value) => {
+                if (value && typeof value === "object" && value.type === "Buffer") {
+                    return Buffer.from(value);
+                }
+                return value;
+            });
+            return obj == null ? {} : obj;
```
tharvik: From the  [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse), `JSON.parse` can't return null.

dynacred/src/Storage.ts
```diff
@@ -32,13 +32,22 @@ export class StorageDB {
...
+        } catch (e) {
+            Log.catch(e, "couldn't parse data", d);
```
tharvik: Catch only `SyntaxError`, to avoid masking other errors.

webapp/src/app/user-data.service.ts
```diff
@@ -36,7 +35,7 @@ export class UserData extends Data {
     async load() {
         Log.lvl1("Loading data from", this.dataFileName);
         const values = await this.storage.getObject(this.dataFileName);
-        if (!values || values === {}) {
+        if (Object.keys(values).length === 0) {
```
tharvik: You can now check for `undefined`.
```suggestion
        if (values === undefined) {
```

PR #167, opened 2019-11-29
Update dynacred for mobile

ineiti:
Adds PollDelete to personhood-rpc.
Contact checks for change in parameter before calling incVersion
Export more classes


dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
 {
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
     "declaration": true,
+    "experimentalDecorators": true,
```
tharvik: I didn't see any decorator use, why do you have it?

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
...
+    "emitDecoratorMetadata": true,
+    "noEmitHelpers": false,
```
tharvik: Why?

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
...
+    "noEmitOnError": true,
+    "allowSyntheticDefaultImports": true,
```
tharvik: Why?

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
 {
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
     "declaration": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
tharvik: decorator related

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
 {
   "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
```
tharvik: Why going to an older version?

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
 {
   "compilerOptions": {
+    "module": "commonjs",
```
tharvik: ES6 as real support for module, no need to explicit `"module"`

dynacred/src/Contact.ts
```diff
@@ -65,7 +65,7 @@ export class Contact {
     }
 
     set alias(a: string) {
-        if (a) {
+        if (a && this.alias !== a) {
```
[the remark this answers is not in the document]
ineiti: The idea here is to only update the version of the `Contact` whenever something changes. So for once I like having a setter that will actually `incVersion` if the attribute changes, and just ignores it if it's the same.

dynacred/tsconfig.json
```diff
@@ -1,17 +1,36 @@
...
+      "dom"
+    ],
+    "baseUrl": ".",
+    "outDir": "./build",
+    "paths": {
+      "~/*": [
+        "src/*"
+      ],
+      "@c4dt/cothority/*": [
+        "node_modules/@c4dt/cothority/*",
```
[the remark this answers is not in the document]
ineiti: yes, until we're using `@c4dt/cothority` correctly without sub-directories, so that I can do `npm link ../../cothority/external/js/cothority`

dynacred/tsconfig.json
ineiti: OK, I'll have another look at these `tsconfig.json` options - it was mainly a manual effort to get `dynacred`, `@dedis/cothority`, `crypto`, `nativescript-nodeify` all working together nicely.

dynacred/src/Contact.ts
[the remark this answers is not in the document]
tharvik: > The idea here is to only update the version of the `Contact` whenever something changes. So for once I like having a setter that will actually `incVersion` if the attribute changes,

I totally agree: `incVersion` in case of change and not if the same.

> and just ignores it if it's the same.

Which is not what's currently implemented, as when passing an empty string, it doesn't get updated.

dynacred/src/Contact.ts
```diff
@@ -511,19 +553,26 @@ export class Contact {
...
-                    const cont = await Contact.fromByzcoin(this.bc, csBuf.slice(c, c + 32), false);
-                    Log.lvl2(`Got contact ${cont.alias} in:`, new Date().getTime() - d.getTime());
-                    this.contactsCache.push(cont);
-                } catch (e) {
-                    Log.error("couldn't get contact - removing contact from the list");
-                    hasFaulty = true;
-                }
+                contactsBuf.push(csBuf.slice(c, c + 32));
             }
+            this.contactsCache = await Promise.all(contactsBuf.map((buf) => {
```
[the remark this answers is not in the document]
ineiti: Hmm - that what is done in the following lines, no?

dynacred/src/Data.ts
```diff
@@ -616,10 +621,32 @@ export class Data {
         this.contacts = this.contacts.filter((u) => !u.equals(nu));
     }
 
+    async updateParties(): Promise<PartyItem[]> {
+        await Promise.all(this.parties.map(async (p) => p.partyInstance.update()));
+        // Move all finalized parties into badges
+        const parties: PartyItem[] = [];
+        this.parties.forEach((p) => {
```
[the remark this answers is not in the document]
ineiti: is it save to work on an array via `map` and assign the result to the same array?

```
isItSave = isItSave.map(s => s.do_something)
```

dynacred/src/personhood-rpc.ts
```diff
@@ -21,13 +21,13 @@ import { Message, Properties } from "protobufjs";
  * depend on anything from dynacred.
  */
 export class PersonhoodRPC {
```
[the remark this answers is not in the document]
ineiti: that's the broken protobuf-over-websockets thingie: if the endpoints are different, you cannot re-use the connection. Also, the `WebSocketConnection` does a cache of existing connections, so it's not a problem.

dynacred/src/Contact.ts
```diff
@@ -453,6 +491,10 @@ export class Contact {
 
     async updateOrConnect(bc: ByzCoinRPC = null, getContacts: boolean = true): Promise<Contact> {
```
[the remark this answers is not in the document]
ineiti: unfortunately the name is very well chosen. It either tries to update the class, or connects to the byzcoin. It's a 2-in-1, and I won't change it... But add a comment.

dynacred/src/Contact.ts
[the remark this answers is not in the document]
tharvik: It is done for `this.contacts`, not for `this.contactsCache`.

dynacred/src/Data.ts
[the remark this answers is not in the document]
tharvik: If you find safety guarantees for JS, tell me :P
But yep, it's safe, it's supposed to create a new array without references to the old one.

dynacred/src/personhood-rpc.ts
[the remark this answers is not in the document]
tharvik: Ho, right...

dynacred/tsconfig.json
ineiti: that's the only one that I need - else `npm run build` fails.

tharvik: Yep, there is a relation between `target` and `module` (I prefer the former as going forward as ES6 as support for module).
https://www.typescriptlang.org/docs/handbook/compiler-options.html

ineiti: As far as I understand, this is how the modules are resolved. So if I remove that, it doesn't find any modules anymore.

dynacred/Makefile
```diff
@@ -1,50 +1,11 @@
...
-	@npm run lint:fix
-
-cothority:
-	git clone --depth=1 https://github.com/${COT}
-
-cothority-pull: cothority
-	cd cothority && git pull
-
-kyber_src: cothority-pull src/lib/kyber
+use_src:
```
tharvik: `.PHONY`

dynacred/Makefile
```diff
@@ -1,50 +1,11 @@
...
+	make -C .. cothority-pull
+	ln -s ../../../cothority/external/js/cothority/src src/lib/cothority
+	ln -s ../../../cothority/external/js/kyber/src src/lib/kyber
 
-src/lib/kyber: cothority/external/js/kyber/src
-	@cp -a $< $@
-	@find src/ -name "*.ts" | xargs perl -pi -e "s:\@dedis/kyber:src/lib/kyber:"
+use_npm:
```
tharvik: `.PHONY`

dynacred/spec/calypso.spec.ts
```diff
@@ -1,8 +1,8 @@
-import DarcInstance from "@dedis/cothority/byzcoin/contracts/darc-instance";
-import { CalypsoReadInstance, CalypsoWriteInstance, OnChainSecretRPC, Write } from "@dedis/cothority/calypso";
-import { Darc, Rule } from "@dedis/cothority/darc";
-import Log from "@dedis/cothority/log";
-import { curve } from "@dedis/kyber";
+import DarcInstance from "@c4dt/cothority/byzcoin/contracts/darc-instance";
```
tharvik: So, what's the default now: do we want dynacred to deps on c4dt's {cothority,kyber} or dedis'?

dynacred/src/Contact.ts
```diff
@@ -52,6 +54,15 @@ export class Contact {
         return svBuf.readInt32LE(0);
     }
 
+    set structVersion(sv: number) {
```
tharvik: `Contact.structVersion = 2` gives `Contact.structVersion === 0`

dynacred/src/Contact.ts
```diff
@@ -80,7 +94,7 @@ export class Contact {
     }
 
     set email(e: string) {
-        if (e) {
+        if (this.email !== e) {
```
tharvik: You're doing the check for empty value in `alias` but not in `email`, `phone` or `url`, wanted?

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
+        const raw = this.credential.getAttribute("1-public", "seedPub");
+        if (raw === undefined) {
+            return undefined;
+        }
+        return Public.fromBuffer(raw);
     }
 
     set seedPublic(pub: Public) {
-        if (pub) {
+        if (this.seedPublic === undefined || !this.seedPublic.equal(pub)) {
```
tharvik:
```suggestion
        if (!pub.equal(this.seedPublic)) {
```

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
-        if (raw === null) {
-            Log.warn(`Contact("${this.alias}").credential has null 1-public/personhood`);
-            return undefined;
-        }
         return Public.fromBuffer(raw);
     }
 
     set personhoodPub(pub: Public) {
-        if (pub) {
+        if (this.personhoodPub === undefined ||
```
tharvik: `undefined` checked by `pub.equal`, you can remove it.

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
     }
 
-    get coinID(): InstanceID {
+    get coinID(): InstanceID | undefined {
         return this.credential.getAttribute("1-public", "coin");
     }
 
     set coinID(id: InstanceID) {
-        if (id) {
+        if (this.coinID === undefined || Buffer.compare(this.coinID, id) !== 0) {
```
tharvik:
```suggestion
        if (!id.equals(this.coinID)) {
```

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
     }
 
-    get ltsID(): InstanceID {
+    get ltsID(): InstanceID | undefined {
         return this.credential.getAttribute("1-config", "ltsID");
     }
 
     set ltsID(id: InstanceID) {
-        if (id) {
+        if (this.ltsID === undefined || !id.equals(this.ltsID)) {
```
tharvik:
```suggestion
        if (!id.equals(this.ltsID)) {
```

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
-    get ltsX(): Point {
+    get ltsX(): Point | undefined {
         const lx = this.credential.getAttribute("1-config", "ltsX");
-        return lx ? PointFactory.fromProto(lx) : null;
+        return lx ? PointFactory.fromProto(lx) : undefined;
     }
 
     set ltsX(X: Point) {
-        if (X) {
+        if (this.ltsX === undefined || !X.equals(this.ltsX)) {
```
tharvik:
```suggestion
        if (!X.equals(this.ltsX)) {
```

dynacred/src/Contact.ts
```diff
@@ -123,82 +135,98 @@ export class Contact {
...
         return this.credential.getAttribute("1-config", "spawner");
     }
 
-    set spawnerID(id: InstanceID) {
-        if (id) {
+    set spawnerID(id: InstanceID | undefined) {
+        if (id === undefined) {
+            throw new Error("spawnerID cannot be undefined");
+        }
+        if (this.spawnerID === undefined || Buffer.compare(this.spawnerID, id) !== 0) {
```
tharvik:
```suggestion
        if (!id.equals(this.spawnerID) {
```

dynacred/src/Contact.ts
```diff
@@ -349,20 +392,20 @@ export class Contact {
...
+    credentialInstance: CredentialsInstance = undefined;
+    darcInstance: DarcInstance = undefined;
+    coinInstance: CoinInstance = undefined;
+    spawnerInstance: SpawnerInstance = undefined;
+    recover: Recover = undefined;
+    calypso: Calypso = undefined;
+    bc: ByzCoinRPC = undefined;
+    contactsCache: Contact[] = undefined;
+    actionsCache: DarcInstance[] = undefined;
+    groupsCache: DarcInstance[] = undefined;
```
tharvik: You don't need to explicit `undefined` assignation, but you can correct the types to be also undefined.

tharvik: Except for `credential`, `recover` and `calypso` as these are definitly assigned in the constructor.

dynacred/src/Contact.ts
```diff
@@ -451,25 +494,43 @@ export class Contact {
         };
     }
 
-    async updateOrConnect(bc: ByzCoinRPC = null, getContacts: boolean = true): Promise<Contact> {
-        if (bc) {
+    async getInstances(): Promise<any> {
```
tharvik:
```suggestion
    async getInstances(): Promise<void> {
```

dynacred/src/Contact.ts
```diff
@@ -481,6 +542,22 @@ export class Contact {
                 case 0:
                     this.credential.setAttribute("1-devices", "initial", this.seedPublic.toBuffer());
                     break;
+                case 1:
+                    const csBufOld = this.credential.getAttribute("1-public", "contacts");
+                    if (csBufOld && csBufOld.length > 0) {
```
tharvik: I think you can replace it with simply `csBufOld !== undefined` but maybe the second check is relevant.

dynacred/src/Contact.ts
```diff
@@ -481,6 +542,22 @@ export class Contact {
...
+                        const csArray: string[] = JSON.parse(csBufOld.toString());
+                        this.contactsCache = csArray.map((c) => Contact.fromObject(c));
+                        for (const c of this.contactsCache) {
+                            await c.updateOrConnect(this.bc, false);
+                            await c.getInstances();
+                            await c.getContacts();
+                        }
+                        this.credential.setAttribute("1-public", "contacts", Buffer.alloc(0));
+                        this.contacts = this.contactsCache;
+                        return;
```
tharvik: Why returning now?

dynacred/src/Contact.ts
```diff
@@ -495,41 +572,33 @@ export class Contact {
...
-                }
-            }
-            if (hasFaulty) {
-                this.contacts = this.contactsCache;
+            const contactsBuf: Buffer[] = [];
+            for (let c = 0; c < csBuf.length; c += iidLen) {
+                contactsBuf.push(csBuf.slice(c, c + iidLen));
             }
+            const cc = await Promise.all(contactsBuf.map((buf) => {
+                return new Promise<Contact>(async (resolve) => {
```
tharvik:
```suggestion
                return new Promise<Contact | undefined>(async (resolve) => {
```

dynacred/src/Contact.ts
```diff
@@ -746,12 +815,12 @@ class Recover {
     }
 
     findTrustee(trustee: InstanceID | Contact): number {
-        if (this.trusteesBuf == null || this.trusteesBuf.length === 0) {
+        if (this.trusteesBuf === undefined || this.trusteesBuf.length === 0) {
```
tharvik: `this.trusteesBuf === undefined` will always be false typewise; but as you're using the length of the buffer to mark missing value, maybe update `get trusteesBuf` to also return `undefined`.

dynacred/src/Data.ts
```diff
@@ -126,7 +128,7 @@ export class Data {
      * @param unrestricted whether the adminDarcID needs evolve_unrestricted
      */
     static async createFirstUser(bc: ByzCoinRPC, adminDarcID: InstanceID, adminKey: Scalar,
-                                 alias: string, unrestricted: boolean = false):
+                                 alias: string, unrestricted?: boolean):
```
tharvik: In this case, you supports three values for `unrestricted`, `true`, `false` and `undefined`. Is it wanted?

Makefile
```diff
@@ -1,15 +1,27 @@
...
+	  done; \
 	  npm run lint:fix; \
 	  cd ..; \
 	done
+	echo ${to} > cot.txt
+
+cothority:
+	git clone https://github.com/${COT}/cothority
+
+cothority-pull: cothority
```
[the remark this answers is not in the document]
ineiti: I know what a PHONY target is, but why is this needed here? `cothority-pull` will never be there, so it is always supposed to be called, no?

TSCONFIG.md
```diff
@@ -0,0 +1,13 @@
+The tsconfig.common.json in this directory is included both by
+`webapp/tsconfig.json` and `dynacred/tsconfig.json` to allow easy
+working with changing cothority- and kyber-sources.
+
+The paths given are resolved with the first found path. So `webapp`
+will always use the dynacred-source, except if packed for distribution
+(which currently is never done). For the cothority- and kyber-sources,
+in normal running it takes the distributed version in
+`node_modules/@(dedis|c4dt)/{cothority,kyber}`.
```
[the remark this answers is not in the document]
ineiti: looked like different scopes to me ;)

dynacred/Makefile
ineiti: why?

dynacred/Makefile
ineiti: why?

dynacred/spec/calypso.spec.ts
ineiti: as soon as I get my stuff into DEDIS we'll do `make cothority_dedis`...

dynacred/src/Contact.ts
ineiti: why?

dynacred/src/Contact.ts
ineiti: yes, alias needs to be non-empty, email, phone, and url can be empty.

dynacred/src/personhood-rpc.ts
```diff
@@ -543,6 +643,44 @@ export class MeetupResponse extends Message<MeetupResponse> {
...
+
+    constructor(props?: Properties<ChallengeCandidate>) {
+        super(props);
+    }
+}
+
+export class ChallengeReply extends Message<ChallengeReply> {
+
+    static register() {
+        registerMessage("ChallengeReply", ChallengeReply);
```
tharvik: So, I've been using the variadic arguments to `registerMessage` to populate the message dependencies; as such, you need solely to register the "edge"-messages, the one visible and relevant to the outside.

```suggestion
        registerMessage("ChallengeReply", ChallengeReply, ChallengeCandidate);
```

dynacred/src/Contact.ts
ineiti: I'm not sure that `pub.equals(undefined)` actually works...

webapp/package.json
```diff
@@ -13,28 +13,27 @@
...
     "angularx-qrcode": "^1.6.4",
-    "core-js": "^2.6.9",
+    "core-js": "^2.6.10",
     "crypto-browserify": "^3.12.0",
     "dexie": "^2.0.4",
     "hammerjs": "^2.0.8",
-    "karma-firefox-launcher": "^1.1.0",
-    "keccak": "^2.0.0",
-    "rxjs": "~6.5.2",
+    "karma-firefox-launcher": "^1.2.0",
```
tharvik: It's testing only, so dev deps.

dynacred/src/Contact.ts
ineiti: OK, I rewrote all of them, supposing for some of them that they can be set to `undefined` or not, depending on the field.

webapp/src/app/app.component.ts
```diff
@@ -27,9 +29,15 @@ export class AppComponent implements OnInit {
     ) {
     }
 
+    logAppend(msg: string, perc: number) {
+        this.log += `\n${msg}`;
```
tharvik: First line is always `\n`

dynacred/src/Contact.ts
ineiti: as there is a `setAttribute` line afterwards, I guess that the field `1-public` / `contacts` stays in the structure, but I'm not sure whether it will have a `buffer.alloc(0)` or `undefined`. Better safe than sorry.

webapp/src/app/app.module.ts
```diff
@@ -48,7 +48,7 @@ import { UserData } from "./user-data.service";
 // This is empty as the UserData is now initialized in the
 // app.component, so that the success can be logged.
 export function loadUserDataConfig(d: UserData) {
-    return () => {};
+    return () => {const a = 2; };
```
tharvik: Why?

dynacred/src/Contact.ts
ineiti: because?

webapp/src/app/user-data.service.ts
```diff
@@ -17,33 +20,34 @@ import Log from "@dedis/cothority/log";
...
-        for (let i = 0; i < 3; i++) {
-            await this.bc.getSignerCounters([]);
-            const url = wsc.getURL();
-            logger(`Fastest node at ${i}/3: ${url}`);
+        logger("Pinging nodes", 10);
+        this.conn = new RosterWSConnection(this.config.roster, StatusRPC.serviceName);
+        this.conn.setParallel(this.config.roster.length);
+        for (let i = 0; i < 5; i++) {
+            await this.conn.send(new StatusRequest(), StatusResponse);
+            const url = this.conn.getURL();
```
tharvik: You can move it outside of the loop.

webapp/src/app/user-data.service.ts
```diff
@@ -17,33 +20,34 @@ import Log from "@dedis/cothority/log";
...
-        wsc.setParallel(5);
-        for (let i = 0; i < 3; i++) {
-            await this.bc.getSignerCounters([]);
-            const url = wsc.getURL();
-            logger(`Fastest node at ${i}/3: ${url}`);
+        logger("Pinging nodes", 10);
+        this.conn = new RosterWSConnection(this.config.roster, StatusRPC.serviceName);
+        this.conn.setParallel(this.config.roster.length);
+        for (let i = 0; i < 5; i++) {
+            await this.conn.send(new StatusRequest(), StatusResponse);
```
tharvik: You have `setParallel > 1` but only doing a single `send` each time.

webapp/src/app/user-data.service.ts
```diff
@@ -17,33 +20,34 @@ import Log from "@dedis/cothority/log";
...
-        const wsc = (<any>this.bc).conn;
-        wsc.setParallel(5);
-        for (let i = 0; i < 3; i++) {
-            await this.bc.getSignerCounters([]);
-            const url = wsc.getURL();
-            logger(`Fastest node at ${i}/3: ${url}`);
+        logger("Pinging nodes", 10);
+        this.conn = new RosterWSConnection(this.config.roster, StatusRPC.serviceName);
+        this.conn.setParallel(this.config.roster.length);
+        for (let i = 0; i < 5; i++) {
```
tharvik: Why `5`?

webapp/src/app/user-data.service.ts
```diff
@@ -17,33 +20,34 @@ import Log from "@dedis/cothority/log";
...
+            await this.conn.send(new StatusRequest(), StatusResponse);
+            const url = this.conn.getURL();
+            logger(`Fastest node at ${i}/3: ${url}`, 20 + i * 15);
```
tharvik:
```suggestion
            logger(`Fastest node at ${i}/5: ${url}`, 20 + i * 15);
```

Makefile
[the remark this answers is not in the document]
tharvik: It's not needed, but I find it more readable (the main target are usually phony).

dynacred/src/Contact.ts
tharvik: Oups, I read the second argument to `svBuf.writeInt32LE` as the value to save.

dynacred/src/Contact.ts
ineiti: made `.*trustee.*` nicer and better.

dynacred/src/Data.ts
ineiti: premature @tharvik optimization...

dynacred/src/Contact.ts
tharvik: You're right, it doesn't support `undefined`.

dynacred/src/Contact.ts
tharvik: So, this doesn't works as `Point.equals` doesn't take `undefined` as argument.

Makefile
```diff
@@ -1,15 +1,27 @@
...
+		npm remove @${from}/$$p; \
+		npm i --save @${to}/$$p; \
+	  done; \
 	  npm run lint:fix; \
 	  cd ..; \
 	done
+	echo ${to} > cot.txt
+
+cothority:
+	git clone https://github.com/${COT}/cothority --depth 10
```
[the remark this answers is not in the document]
ineiti: random value to have _some_ history.

dynacred/src/Data.ts
```diff
@@ -334,58 +337,59 @@ export class Data {
...
+            references: this.references,
             ropascis: [] as any,
         };
         if (this.bc) {
             v.bcRoster = this.bc.getConfig().roster.toJSON();
             v.bcID = this.bc.getGenesis().computeHash();
-            v.parties = this.parties ? this.parties.map((p) => p.toObject()) : null;
-            v.badges = this.badges ? this.badges.map((b) => b.toObject()) : null;
-            v.ropascis = this.ropascis ? this.ropascis.map((rps) => rps.toBytes()) : null;
+            v.parties = this.parties ? this.parties.map((p) => p.toObject()) : undefined;
```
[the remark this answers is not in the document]
ineiti: ???

webapp/src/app/app.module.ts
ineiti: cannot be empty else tslint complains...

webapp/src/app/user-data.service.ts
ineiti: no, because it changes - `this.conn.getURL()` returns the first node, which is re-arranged to be the fastest one after each `conn.send` call.

webapp/src/app/user-data.service.ts
ineiti: doesn't matter, the `setParallel` defines how many identitcal messages are sent to the nodes (in case one or the other is down / censoring).

webapp/src/app/user-data.service.ts
ineiti: random value.

dynacred/src/Data.ts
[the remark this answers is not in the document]
tharvik: Actively check for `undefined`, and even better, we can use optional chaining.

```suggestion
            v.parties = this.parties?.map((p) => p.toObject());
```

webapp/src/app/app.module.ts
tharvik: Does removing `loadUserDataConfig` and the block using it in `providers` (so that `providers: [UserData]`) works?

ineiti: no, because `useFactory` is needed in a provider and must return a function.

dynacred/src/Data.ts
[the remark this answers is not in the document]
ineiti: doesn't seem to be supported in chrome...

PR #175, opened 2019-12-09
Simple explorer

ineiti:
Adds an explorer to omniledger with links to other instances.

Instead of waiting for the full phonebook, already merging this now, so that it can be used with an upcoming update for the personhood-contract.

### DIFF
--- a/webapp/src/app/app-routing.module.ts
+++ b/webapp/src/app/app-routing.module.ts
@@ -12,6 +12,7 @@ import { C4dtComponent } from "./c4dt/c4dt.component";
 import { PartnerComponent } from "./c4dt/partner/partner.component";
 import { ProfileComponent } from "./c4dt/profile/profile.component";
 import { WelcomeComponent } from "./c4dt/welcome/welcome.component";
+import { ExplorerComponent } from "./explorer/explorer.component";
 import { NewuserComponent } from "./newuser/newuser.component";
 import { DeviceComponent } from "./register/device/device.component";
 import { RegisterComponent } from "./register/register.component";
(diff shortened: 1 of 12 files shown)


dynacred/src/Config.ts
```diff
@@ -33,6 +33,13 @@ export class Config {
             return Buffer.from(field, "hex");
         };
 
+        const asBool = (field: any): boolean => {
```
tharvik: Hum, TOML supports boolean directly

webapp/src/app/app-routing.module.ts
```diff
@@ -1,13 +1,14 @@
 import { CommonModule } from "@angular/common";
 import { NgModule } from "@angular/core";
 import { RouterModule, Routes } from "@angular/router";
+import { ExplorerComponent } from "src/app/explorer/explorer.component";
```
tharvik:
```suggestion
import { ExplorerComponent } from "./explorer/explorer.component";
```

webapp/src/app/explorer/darc/darc.component.ts
```diff
@@ -0,0 +1,31 @@
...
+type ExprValue = [string, string];
+type Rule = [string, ExprValue[]];
+
+@Component({
+  selector: "app-darc",
+  templateUrl: "./darc.component.html",
+})
+export class DarcComponent implements OnChanges {
+
+  rules: Rule[];
```
tharvik:
```suggestion
  rules: Rule[] | undefined;
```

webapp/src/app/explorer/explorer.component.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    constructor(
+        private readonly route: ActivatedRoute,
+        private readonly uData: UserData,
+    ) {}
+
+    async ngOnInit() {
+        this.route.paramMap.subscribe(async (params) => {
+            Log.print("new params:", params);
+            Log.print("new params:", params.get("id"));
```
tharvik: debug?

dynacred/src/Config.ts
ineiti: debugging...

PR #185, opened 2019-12-12
Moving back to dedis

ineiti:
Now that all PRs have been integrated upstream, and that a new version is published on
npm, move the dependencies to upstream.


webapp/Makefile
```diff
@@ -1,12 +1,16 @@
+deps:
+	npm ci
+	npm link ../dynacred
+
 none:
 	@echo "Not breaking if you don't tell me what to do"
 
-use_src:
+src:
 	make -C ../dynacred use_src
```
tharvik: You renamed `use_src` to `src`.

webapp/Makefile
```diff
@@ -1,12 +1,16 @@
...
 	ln -s ../../../cothority/external/js/cothority/src src/lib/cothority
 	ln -s ../../../cothority/external/js/kyber/src src/lib/kyber
 
-use_npm:
+npm:
 	make -C ../dynacred use_npm
```
tharvik: Renamed also.

ineiti:
In favor for #191

PR #189, opened 2019-12-13
Update Makefiles and helpers

ineiti:
This PR makes sure that all Makefiles can be used as they are supposed to be,
in order to:
- dynacred, mobile: build or prepare the directory
- conode, webapp: run the app in the directory
- dynacred, mobile, webapp: switch between libraries as source or as npms

### DIFF
--- a/conode/Makefile
+++ b/conode/Makefile
@@ -39,17 +39,16 @@ docker: build/Dockerfile cothority-pull
 	docker build -t $(IMAGE_NAME):$(TAG) -f build/Dockerfile build
 	docker tag $(IMAGE_NAME):$(TAG) $(IMAGE_NAME):dev
 
-update_assets:
-	docker rm -f $(CONTAINER) || echo nothing to stop
-	echo "ByzCoinID = \"$$( grep ByzCoinID variables.txt | sed -e 's/.* //' )\"" > $(WEBAPP_ASSET)
-	echo >> $(WEBAPP_ASSET)
-	cat build/conodes/public.toml >> $(WEBAPP_ASSET)
+$(WEPAPP_ASSET): variables.txt build/conodes/public.toml
+	 echo "ByzCoinID = \"$$( grep ByzCoinID $< | sed -e 's/.* //' )\"" > $@
+	 echo >> $@
+	 cat $(lastword $^) >> $@
 
 docker_run: docker update_assets
 	docker run --rm -ti -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
 
 docker_run_travis: docker
-	docker run --rm -t -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
+	docker run --rm -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
 
 docker_wait:
 	@for port in $$( seq 7771 2 7777 ); do \
--- a/conode/Makefile
+++ b/conode/Makefile
@@ -5,10 +5,11 @@ WEBAPP_ASSET := ../webapp/src/assets/config.toml
 
 all: docker
 
-.PHONY: conode
+.PHONY: cothority-pull conode newdb docker docker_run docker_wait docker_stop docker_start docker_clean clean
 
-conode:
-	make -C .. cothority-pull
+.DEFAULT_GOAL := docker_run
+
+utils: cothority-pull
 	@echo "Compiling binaries"; \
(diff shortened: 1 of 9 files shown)


conode/Makefile
```diff
@@ -5,18 +5,22 @@ WEBAPP_ASSET := ../webapp/src/assets/config.toml
...
 all: docker
 
-.PHONY: conode
+.PHONY: cothority-pull conode newdb docker docker_run docker_wait docker_stop docker_start docker_clean clean
 
-conode:
-	make -C .. cothority-pull
+.DEFAULT_GOAL := docker_run
+
+utils: cothority-pull
```
tharvik: To avoid uneeded rebuild (and to use the file deps system that make provide), you can split the target:
```make
.PHONY: utils
utils: build/bcadmin build/csadmin build/phapp
build/bcadmin: $(wildcard ../conode/*.go ../conode/*/*.go)
        go build -o $@ ../conode/byzcoin/bcadmin
# usw...
```

conode/Makefile
```diff
@@ -29,20 +33,22 @@ newdb: conode
...
 
 # Use this target to build from local source
-docker: build/Dockerfile conode
+docker: build/Dockerfile cothority-pull
 	@export GO111MODULE=on GOOS=linux GOARCH=amd64; \
-	go build -o build/conode; \
-	cd ../cothority && \
-		go build -o ../conode/build/bcadmin ./byzcoin/bcadmin && \
-		go build -o ../conode/build/csadmin ./calypso/csadmin
+	go build -o build/conode
```
tharvik: You can extract this build.

conode/Makefile
```diff
@@ -29,20 +33,22 @@ newdb: conode
...
 	docker build -t $(IMAGE_NAME):$(TAG) -f build/Dockerfile build
 	docker tag $(IMAGE_NAME):$(TAG) $(IMAGE_NAME):dev
 
-docker_run: docker
+update_assets:
```
tharvik: Extractable, also, why stopping the docker in this target?

```make
$(WEPAPP_ASSET): variables.txt build/conodes/public.toml
        echo "ByzCoinID = \"$$( grep ByzCoinID $< | sed -e 's/.* //' )\"" > $@
        echo >> $@
        cat $(lastword $^) >> $@
```

conode/Makefile
```diff
@@ -29,20 +33,22 @@ newdb: conode
...
 	docker rm -f $(CONTAINER) || echo nothing to stop
 	echo "ByzCoinID = \"$$( grep ByzCoinID variables.txt | sed -e 's/.* //' )\"" > $(WEBAPP_ASSET)
 	echo >> $(WEBAPP_ASSET)
 	cat build/conodes/public.toml >> $(WEBAPP_ASSET)
+
+docker_run: docker update_assets
+	docker run --rm -ti -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
+
+docker_run_travis: docker
 	docker run --rm -t -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
```
tharvik: Usually, you don't need a TTY (`-t`) if you don't read stdin.

conode/go.mod
```diff
@@ -12,4 +12,6 @@ require (
 	google.golang.org/appengine v1.6.5 // indirect
 )
 
+replace go.dedis.ch/cothority/v3 => ../cothority
```
tharvik: Debug?

conode/Makefile
ineiti: no, because once its for the target machine, once it's for the Dockerfile, which, on my computer, won't like the mac-binary.

conode/Makefile
ineiti: does it still print to stdout without the `-t`?

conode/go.mod
ineiti: No debug this time, I want it to use the cothority checked out locally. Most of the time it's not the same as master...

conode/Makefile
ineiti: I prefer depending on go's vision of which files are new and whether it needs to re-compile. If I depend on these dependencies, I'm sure somebody (perhaps me) will stay with an old binary because a dependency wasn't included and then wonder why the binary is still the old one...

conode/Makefile
tharvik: Yes

conode/Makefile
ineiti: looks good, thanks.

PR #196, opened 2019-12-19
Adds Contact.getDevices

ineiti:
This makes it possible to easily retrieve the devices of a contact.
Useful if you want to use the public key(s) of a contact.

### DIFF
--- a/dynacred/src/Contact.ts
+++ b/dynacred/src/Contact.ts
@@ -792,13 +792,13 @@ class Device {
         const d = await DarcInstance.fromByzcoin(rpc, darcID);
         return new Device(d.darc);
     }
-    pubKey: Public;
+    readonly pubKey: Public;
 
     /**
      * Constructor verifies the given darc is somewhat compatible with a device.
      * @param darc
      */
-    constructor(public darc: Darc) {
+    constructor(readonly darc: Darc) {
         const ids = darc.rules.getRule(Darc.ruleSign).getIdentities();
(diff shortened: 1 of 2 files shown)


dynacred/src/Contact.ts
```diff
@@ -766,6 +777,40 @@ interface IHasAlias {
...
+        const d = await DarcInstance.fromByzcoin(rpc, darcID);
+        return new Device(d.darc);
+    }
+    pubKey: Public;
+
+    /**
+     * Constructor verifies the given darc is somewhat compatible with a device.
+     * @param darc
+     */
+    constructor(public darc: Darc) {
```
tharvik:
```suggestion
    constructor(public readonly darc: Darc) {
```

dynacred/src/Contact.ts
```diff
@@ -766,6 +777,40 @@ interface IHasAlias {
...
+    /**
+     * Fetch a new device from Byzcoin given the darcID of the signer.
+     * @param rpc
+     * @param darcID
+     */
+    static async fromByzcoin(rpc: ByzCoinRPC, darcID: InstanceID): Promise<Device> {
+        const d = await DarcInstance.fromByzcoin(rpc, darcID);
+        return new Device(d.darc);
+    }
+    pubKey: Public;
```
tharvik:
```suggestion
    public readonly pubKey: Public;
```

PR #198, opened 2019-12-20
Managing devices

### LINKED ISSUE #10: webserver: allow to change device name
Currently, the first device is called "initial", but that's not very descriptive for the user. Naming is an evolving notion so renaming can be very useful.

### LINKED ISSUE #176: Confirm deletion of device
When a user wants to remove a device, there needs to be a confirmation dialog.

### LINKED ISSUE #178: Overwrite device with same name
A user can create a device with same name which overwrites existing device.

TODO: refuse to create a device with an existing name.

ineiti:
Closes #176
Closes #178
Closes #10


webapp/src/app/admin/devices/devices.component.ts
```diff
@@ -53,22 +55,31 @@ export class DevicesComponent implements OnInit {
...
-        await showTransactions(this.dialog, "Deleting device " + device.name,
-            async (progress: TProgress) => {
-                progress(30, "Deleting Device");
-                await this.uData.contact.deleteDevice(device.name);
-                progress(60, "Updating Device List");
-                await this.uData.contact.sendUpdate();
-                progress(-90, "Fetching all devices");
-                this.updateDevices();
-            });
+        if (await showDialogOKC(this.dialog, `Deleting device`,
```
tharvik: Extracting the condition to a value makes it more readable.

webapp/src/app/admin/devices/devices.component.ts
```diff
@@ -53,22 +55,31 @@ export class DevicesComponent implements OnInit {
...
+                });
+        }
+
     }
 
-    async add() {
+    async addDevice() {
         const ac = this.dialog.open(DeviceAddComponent);
         ac.afterClosed().subscribe(async (result) => {
             if (result) {
```
tharvik:
```suggestion
             if (result === undefined || result === "") {
```

webapp/src/app/admin/devices/devices.component.ts
```diff
@@ -84,11 +95,40 @@ export class DevicesComponent implements OnInit {
...
+    async renameDevice(dev: Device, typeStr: string) {
+        const ac = this.dialog.open(RenameComponent, {data: {name: dev.name, typeStr}});
+        ac.afterClosed().subscribe(async (result) => {
+            if (result === undefined || result === "") {
+                return;
+            }
+            if (this.devices.concat(this.recovery).find((d) => d.name === result)) {
+                await showDialogInfo(this.dialog, "Duplicate name", `The ${typeStr}-name ${result} already exists`,
+                    "Change");
+                return this.addDevice();
```
tharvik:
```suggestion
                return this.renameDevice();
```

webapp/src/app/admin/devices/devices.component.ts
```diff
@@ -84,11 +95,40 @@ export class DevicesComponent implements OnInit {
                         });
                 const url = window.location.protocol + "//" + window.location.host +
                     this.location.prepareExternalUrl(device);
-                this.dialog.open(DeviceShowComponent, {data: url});
+                this.dialog.open(ShowComponent, {data: url});
             }
         });
     }
 
+    async renameDevice(dev: Device, typeStr: string) {
```
tharvik: `typeStr` is used mainly to check if it is handling a recovery device or not. Maybe renaming it to `isRecovery: boolean` to show that and avoid external invalid calls.

webapp/src/app/admin/devices/devices.component.ts
```diff
@@ -153,23 +200,43 @@ export class DevicesComponent implements OnInit {
...
 }
 
 @Component({
-    selector: "device-show",
-    templateUrl: "device-show.html",
+    selector: "show",
+    templateUrl: "show.html",
 })
-export class DeviceShowComponent {
+export class ShowComponent {
```
tharvik: Nice! I find it more readable this way (keeping related items together, short names, ...)
Now, to be picky, why then the `rename` to `renameDevice` transformation (amongs other)?

webapp/src/app/admin/devices/devices.component.ts
ineiti: Please take care when proposing these changes. I saw several inverted suggestions from your part, and one took me quite some time to discover afterwards...

webapp/src/app/admin/devices/devices.component.ts
ineiti: Doesn't look like a boolean to me. It could be an `enum`.

webapp/src/app/admin/devices/devices.component.ts
ineiti: Good catch. `rename` handles both devices and recovery, so it should not be `renameDevice`. The other methods only handle one case, so I keep them like that.

webapp/src/app/admin/devices/devices.component.ts
ineiti: Missing arguments in `renameDevice`.

webapp/src/app/admin/devices/devices.component.ts
tharvik: You kept a string type; at least, use a [string literal type](https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types) if you really want to use the string valuespace.

PR #200, opened 2020-01-06
nicer login screen

### LINKED ISSUE #199: Login service feedback
The login service in omniledger should not show the transaction-dialog when checking the authorization of the user.

ineiti:
Closes #199


ineiti:
> It's quite cumbersome to have to do a mental diff when the real changes are only capitalising a few phrases and unwrapping three lines, which would amount to at most 8 lines diff, not 40... Can you clean your diffs beforehand?

We should in fact add a linter to the html files. Currently I'm relying on IntelliJ's integrated linter, which pointed out that the `<p> <p> </p> </p>` construct you used in the html is illegal, so I replaced that with a `<div> <p> </p> </div>`.

tharvik:
> We should in fact add a linter to the html files.

Agree.

> Currently I'm relying on IntelliJ's integrated linter, which pointed out that the `<p> <p> </p> </p>` construct you used in the html is illegal, so I replaced that with a `<div> <p> </p> </div>`.

As I see it, IntelliJ is also formatting differently/more than tslint, is there a way to avoid that? When you edit a file I modified, there is usually many cosmetic changes.

Issue #203, opened 2020-01-17
Lots of connections when logging in


ineiti:
When connecting through CAS, there are a lot of connection requests. For a new user, it didn't work to log in afterwards. Not sure this is related or not.

ineiti:
See real problem in #208

PR #209, opened 2020-01-17
Real fix for endless loop and long block loading …

### LINKED ISSUE #206: Save and load latest block
Currently the latest block is not stored when updated. This should happen, so that login is faster.

### LINKED ISSUE #208: Endless loop in updating blocks
The real problem behind #203 is the bug in dedis/cothority#2181 that allows it to loop infinitely when looking for a block-update.

ineiti:
This fixes the real problems behind #203. It also stores the latest block
locally and starts to update from that one, thus making login much
faster (~1sec instead of ~5sec).

Closes #206
Closes #208

### DIFF
--- a/webapp/src/app/user-data.service.ts
+++ b/webapp/src/app/user-data.service.ts
@@ -26,7 +26,7 @@ export class UserData extends Data {
     private readonly dbLatest = "latest_skipblock";
     // This is the hardcoded block at 0x6000, supposed to have higher forward-links. Once 0x8000 is created,
     // this will be updated.
-    private readonly id0x6000 = "3781100c76ab3e6243da881036372387f8237c59cedd27fa0f556f71dc2dff48";
+    private readonly id0x6000 = Buffer.from("3781100c76ab3e6243da881036372387f8237c59cedd27fa0f556f71dc2dff48", "hex");
 
     constructor() {
         super(undefined); // poison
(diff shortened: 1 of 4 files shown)


webapp/src/app/user-data.service.ts
```diff
@@ -22,6 +23,10 @@ export class UserData extends Data {
     bc: ByzCoinRPC;
     config: Config;
     conn: IConnection;
+    private readonly dbLatest = "latest_skipblock";
+    // This is the hardcoded block at 0x6000, supposed to have higher forward-links. Once 0x8000 is created,
+    // this will be updated.
+    private readonly id0x6000 = "3781100c76ab3e6243da881036372387f8237c59cedd27fa0f556f71dc2dff48";
```
tharvik:
```suggestion
    private readonly id0x6000 = Buffer.from("3781100c76ab3e6243da881036372387f8237c59cedd27fa0f556f71dc2dff48", "hex");
```

webapp/src/app/user-data.service.ts
```diff
@@ -22,6 +23,10 @@ export class UserData extends Data {
     bc: ByzCoinRPC;
     config: Config;
     conn: IConnection;
+    private readonly dbLatest = "latest_skipblock";
```
tharvik: Better name
```suggestion
    private readonly stroageKeyForLatest = "latest_skipblock";
```

PR #224, opened 2020-01-29
Also allow c4dt/kyber for crypto-browserify

ineiti:
The completely black-magick patch for nodeify needs an addition for the
@c4dt/kyber module.


ineiti:
Heh - look at the `Files changed` - github doesn't like showing changes to a patch-file ;)

PR #226, opened 2020-02-03
Adding signature-string to forwardlinks

ineiti:
Simple `xxx-xxx` string to indicate signatures. Only works for 32 nodes...

Also contains fix for new personhood dir-structure in `dedis/cothority#3.4.1`


webapp/src/app/bcviewer/bcviewer.component.ts
```diff
@@ -188,13 +188,27 @@ class LinkBlock {
     index: number;
     height: number;
     maxHeight: number;
+    sign: string;
```
tharvik: It's not definitley assigned in the constructor (ie it can be `undefined`): either change it to `string | undefined` or set it.

webapp/src/app/bcviewer/bcviewer.component.ts
```diff
@@ -188,13 +188,27 @@ class LinkBlock {
...
-            this.id = (link as ForwardLink).to;
+            const l = (link as ForwardLink);
+            this.id = l.to;
+            if (sbNow !== undefined) {
+                this.sign = "";
+                // TODO: extend to more than 32 nodes
+                const maskBuf = Buffer.alloc(4);
+                l.signature.getMask().copy(maskBuf);
+                const mask = Buffer.from(maskBuf.reverse()).readInt32BE(0);
+                const roster = l.newRoster || sbNow.roster;
```
tharvik: In this case, it would be better to use the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_Coalescing_Operator) but that's not yet stable. It has the same kind of issue than using plain `if` to check for `undefined`: the left operand is casted to a boolean and equated to `false`, which would be the case for `undefined` and `null`, but also for `NaN`, `0`, `""`, ...

So, for now, better explicilty check for `undefined`
```suggestion
                const roster = l.newRoster !== undefined ? l.newRoster : sbNow.roster;
```

webapp/src/app/bcviewer/show-block.html
```diff
@@ -27,7 +31,7 @@ <h1>Block {{data.sb.index}}</h1>
                         </div>
                         <div fxFlex="1 1 20em">ForwardLinks:
                             <span *ngFor="let forward of data.forwardLinks" (click)="goBlock(forward)" class="link">
-                                {{forward.index}}</span>
+                                {{forward.index}}/{{forward.sign}}</span>
```
tharvik: Maybe disable it if `forward.sign === undefined`?

ineiti: Is there a code-path where this can happen? All forward-links should have a `sign` field.

webapp/src/app/bcviewer/bcviewer.component.ts
ineiti: set to `""` by default.

webapp/src/app/bcviewer/bcviewer.component.ts
ineiti: hmm - as `newRoster` is either undefined or of type `Roster`, I don't see the problem here. In the definition in `skipblock.ts` it should be `Roster | undefined`.

webapp/src/app/bcviewer/show-block.html
tharvik: Correct, it will always be defined; for the empty string case, it will show as `$index/`, it can be weird to have a trailing slash, but that's a matter of taste.

webapp/src/app/bcviewer/bcviewer.component.ts
tharvik: > In the definition in `skipblock.ts` it should be `Roster | undefined`.

I only trust explicit types :P but yeah, that won't realistically happen.

webapp/src/app/bcviewer/show-block.html
ineiti: well, not as weird as having a forward link without signatures!

PR #234, opened 2020-02-28
webapp/cas: use common config

tharvik:
* cas was using two config, which got out of sync
* now, it uses the same config, so adding a parsing for it

### DIFF
--- a/webapp/src/app/api/v0/cas/config.service.ts
+++ b/webapp/src/app/api/v0/cas/config.service.ts
@@ -1,0 +1,150 @@
+import crypto from "crypto-browserify";
+import { List, Map } from "immutable";
+import toml from "toml";
+
+import { Injectable } from "@angular/core";
+
+import { InstanceID } from "@dedis/cothority/byzcoin";
+
+// Action represents a DARC with its CoinInstance.
+export class Action {
+    constructor(
+        readonly darc: InstanceID,
+        readonly coin: InstanceID,
+    ) {}
+}
+
+function getField(obj: object, name: string): unknown {
+    if (!(name in obj)) {
+        throw new Error(`field "${name}" not found in config`);
+    }
+    return obj[name];
+}
+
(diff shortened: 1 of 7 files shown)


webapp/src/app/api/v0/cas/config.service.ts
```diff
@@ -0,0 +1,147 @@
...
+import { InstanceID } from "@dedis/cothority/byzcoin";
+
+export class Action {
+    constructor(
+        readonly darc: InstanceID,
+        readonly coin: InstanceID,
+    ) {}
+}
+
+const getField = (obj: object, name: string): unknown => {
```
ineiti: Why are these defined as `const` and not as `function`?

webapp/src/app/api/v0/cas/config.service.ts
```diff
@@ -0,0 +1,147 @@
+import crypto from "crypto";
+import { List, Map } from "immutable";
+import toml from "toml";
+
+import { Injectable } from "@angular/core";
+
+import { InstanceID } from "@dedis/cothority/byzcoin";
+
+export class Action {
```
ineiti: I'd take a small comment here (and other exported symbols) for two kind of persons:
- you in 2 months
- everybody else

webapp/src/app/api/v0/cas/login/login.component.ts
```diff
@@ -1,4 +1,4 @@
-import crypto from "crypto-browserify";
+import crypto from "crypto";
```
ineiti: Please use `crypto-browserify` for the moment. Else we have two different sources of crypto, and I don't want to chase those bugs.

webapp/src/app/api/v0/cas/config.service.ts
tharvik: Cruft from a previous iteration, fixing.

PR #243, opened 2020-05-12
CAS: fix headless test

tharvik:
* bump to go 1.14
* kill browser when exiting
* better user injector, which actually wait for the end of the script
* wordpress clicker checks for correct content, and work on small/headless screen


webapp/cas/cmd/cas-test/main.go
```diff
@@ -89,24 +95,33 @@ func getUserData(path string) (string, error) {
...
+				createObjectStore(objectStoreName, {keyPath: "name"}).
+				createIndex("buffer", "buffer", {unique: false, multiEntry: false})
+		request.onsuccess = (event) =>
+			event.target.result.transaction(objectStoreName, "readwrite").
+				objectStore(objectStoreName).
+				put({name: "dynasent"}).
+				onsuccess = resolve
+	})
 
 	Promise.all([dynasentDone, dbnamesDone]).then(() => {
```
ineiti, line 126: Is it possible to bubble up an error to go if you add `catch` here?

webapp/cas/cmd/cas-test/main.go
```diff
@@ -122,7 +137,13 @@ func injectUser(page *agouti.Page, userData string) error {
 		return err
 	}
 
-	page.FindByID("run-script-done")
+	count, err := page.FindByID("run-script-done").Count()
+	if err != nil {
+		return err
+	}
+	if count != 1 {
+		return errors.New("script not run to completion")
```
ineiti, line 145:
```suggestion
		return errors.New("script didn't complete")
```

webapp/cas/cmd/cas-test/main.go
```diff
@@ -122,7 +137,13 @@ func injectUser(page *agouti.Page, userData string) error {
 		return err
 	}
 
-	page.FindByID("run-script-done")
+	count, err := page.FindByID("run-script-done").Count()
+	if err != nil {
+		return err
```
ineiti, line 142:
```suggestion
		return fmt.Errorf("couldn't get run-script-done tag: %v", err)
```

webapp/cas/cmd/cas-test/main.go
tharvik: I don't think there is with `github.com/sclevine/agouti.Page.RunScript`: as all the processing is done in async, the script actually returns `undefined` and I use this `document.createElement` trick to actually wait for the end.
Hum, but I can actually return a potential error message _inside_ the `span`, that's a nice touch :)

ineiti:
Github doesn't see Travis being happy...

tharvik:
Yep, it does that from times to times, pretty annoying..

PR #279, opened 2020-12-01
Warn if storage cannot be used

ineiti:
If the browser is in private mode or refuses to use cookies, it doesn't
allow to store data. But this feature is needed by OmniLedger login.

This PR shows an explanation in that case and points to a c4dt.org
article with more information.

closes dedis/cothority#2415

### DIFF
--- a/webapp/src/app/byz-coin.service.ts
+++ b/webapp/src/app/byz-coin.service.ts
@@ -9,14 +9,13 @@ import Dexie from "dexie";
 import { Fetcher, User } from "dynacred";
 import { Config } from "src/lib/config";
 
-function dbError(e: Error): Error {
-    return new Error("Couldn't open storage of browser.<br>This is probably due to one" +
-        " of:<br>" +
-        "- Using private browsing mode<br>" +
-        "- Setting your browser to not remember history<br>" +
-        "For more information, go to <a" +
-        " href='https://www.c4dt.org/article/failing-omniledger-login/' target='other'>OmniLedger login problems</a>" +
-        "<br><br>System error message: " + e.toString());
+export class DBError implements Error {
+  readonly name = "DBError";
+  readonly message: string;
+
+  constructor(error: Error)  {
+    this.message = error.message;
+  }
 }
 
 @Injectable({
@@ -56,7 +55,7 @@ export class ByzCoinService extends Fetcher {
         try {
             this.db = new StorageDB();
         } catch (e) {
-            throw dbError(e);
+            throw new DBError(e);
         }
         const sc = new SkipchainRPC(this.conn);
         // @ts-ignore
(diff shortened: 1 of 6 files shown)


webapp/src/app/byz-coin.service.ts
```diff
@@ -9,6 +9,16 @@ import Dexie from "dexie";
...
 import { Config } from "src/lib/config";
 
+function dbError(e: Error): Error {
+    return new Error("Couldn't open storage of browser.<br>This is probably due to one" +
+        " of:<br>" +
+        "- Using private browsing mode<br>" +
+        "- Setting your browser to not remember history<br>" +
+        "For more information, go to <a" +
+        " href='https://www.c4dt.org/article/failing-omniledger-login/' target='other'>OmniLedger login problems</a>" +
+        "<br><br>System error message: " + e.toString());
```
tharvik, line 19: It is not recommanded to generate HTML via typescript, as someone can insert random code here by manipulating the returned error. You should create a dedicated component which only accept an error which will be safely escaped by angular when interpolated.

tharvik (changes requested):

webapp/src/app/byz-coin.service.ts
ineiti: Can't I simply escape the `e.toString()` with https://stackoverflow.com/questions/24816/escaping-html-strings-with-jquery ?

tharvik: Maybe. Usually, writing yourself your own escaping library is very prone to bugs (as the mottos of "don't do your own crypto" or "don't trust your own parser"), and there is facilities already existing inside angular so why not use it direclty?

ineiti: 'cause I'm lazy? Anyway, I found a third one: `don't write your own testing harness`...

ineiti: In fact the error lies even deeper: `showDialogInfo` accepts HTML code, so there must be many parts where the error might be used to inject HTML code... So I just added the `encodeURI` for the `e.toString()` here. OK?

tharvik: That won't solve the security issue, only complexify a bit the payload needed to trigger it. The secure way is to create an independent component for each message.

If you really want to keep this pattern, look at https://angular.io/guide/security#direct-use-of-the-dom-apis-and-explicit-sanitization-calls on how to use angular's calls to make it secure.

ineiti: Can you please create a commit to implement what you propose?

tharvik: Here you go, I pushed it to this branch.
I found the root issue that allowed it to work. In `lib/dialog-okcancel.html`, it uses `innerHTML` which is a huge security risk as it allow trivial XSS. As only the `dbError` calls where using HTML formating, it removed it.

ineiti: Cool - thanks a lot for taking care of this. I gotta admit I'm a bit disappointed by the slow user uptake, so I don't really care if we have a possible XSS attack from typescript-errors. Kudos to the one exploiting it ;)

But of course it's much nicer like that! Great!

tharvik (approved):

PR #295, opened 2021-03-16
Add dark theme to omniledger login

ineiti:
This patch changes the css-based angular project to an scss-based one.
This allows a simple setup of a light and a dark theme.
Using the @media selector, the dark theme is activated if it's the system one.


ineiti:
Hmm, I changed it, but it's not supposed to be empty :(

ineiti:
Now it shows it correctly as a binary file.

PR #301, opened 2021-04-15
re-enable using the source

ineiti:


Makefile
```diff
@@ -23,8 +23,15 @@ swap:
...
 	git clone https://github.com/${COT}/cothority --depth 1
 
-cothority-pull: cothority
-	cd cothority && git pull
+src: cothority
+	( cd cothority && git pull && \
+	  cd external/js/kyber && npm ci && \
+	  cd ../cothority && npm ci )
+	( cd dynacred && npm ci -D )
+	( cd webapp && npm ci -D )
```
tharvik, line 31: By default, each line is executed in a different shell, no need to parenthesize it. And I never saw the `-D` in `npm ci`, what's its use?
```suggestion
	cd cothority && git pull
	cd cothority/external/js/kyber && npm ci
	cd cothority/external/js/cothority && npm ci
	cd dynacred && npm ci
	cd webapp && npm ci
```

ineiti: I thought that `-D` installs also the devDependencies. But it seems that `npm ci` installs them anyway...

tharvik (approved):

PR #303, opened 2021-04-19
Ng 10

ineiti:
Update to angular@10.
Angular@11 does give strange errors while compiling. That's for another day...


.travis.yml
```diff
@@ -23,31 +23,34 @@ jobs:
...
+      node_js: "12"
       install:
         - gimme 1.12.4
         - . $HOME/.gimme/envs/go1.12.4.env
         - ( cd dynacred && npm ci )
         - cd webapp && npm ci
       before_script: npm run lint
       script:
-        - npm run build --prod
+        - npm run build
```
tharvik, line 48: as what is deployed is compiled as `npm run -- ng build --prod --base-href /omniledger/ --aot`, CI'ing with the `--prod --aot` is safer.
btw, remember me why we run a `bin/update.sh` script on the server and not rsyncing what was build?

ineiti: I was trying to get the travis-time down from 15 minutes. On my computer the `build:prod` takes twice as long as `build`. But for travis it doesn't seem to make a difference :(

With regard to the `update.sh` script, if you think it's important, please open a new issue.

so I can put `build:prod` back in.

tharvik: > But for travis it doesn't seem to make a difference :(

Then you can put it back as it doesn't change a thing :)

> With regard to the `update.sh` script, if you think it's important, please open a new issue.

https://github.com/c4dt/omniledger/issues/304

tharvik (approved):

PR #305, opened 2021-05-04
Updated the READMEs

LucasTrg:


LucasTrg:
I added a paragraph to describe how the users can simply build using DEDIS' test-network of conodes
I was not sure yet whether or not I should delete the redundant ```Development Server``` paragraph in webapp/README.md, as it seems a bit superfluous. I believe it should be merged with the ```Running it locally``` one.
I made sure to specify where the user should cd as I was a bit confused by all the differents makefiles.

webapp/README.md
```diff
@@ -27,10 +36,13 @@ at this stage of development, you most probably will want to / have to work with
 code:
 
 ```bash
-make -C conode docker
+mv src/assets/config.toml src/assets/backup_config.toml
```
ineiti, line 39: can be removed

PR #306, opened 2021-05-04
Fixes the build process in conode Makefile

LucasTrg:


ineiti:
I think you can close that one in favour of #309

PR #309, opened 2021-05-11
Removing the creation of its own docker

ineiti:
This PR makes the omniledger/conode use the c4dt/byzcoin docker to create a local testing node array.
It also updates correctly the variables in the first-time registration in the webapp.


ineiti:
So it was a `0600` permission on the `*.db` files that made it impossible for `docker build` to copy the files...

conode/Makefile
```diff
@@ -1,54 +1,55 @@
...
 
-utils: cothority-pull
-	@echo "Compiling binaries"; \
-	go build -o build/conode; \
-	cd ../cothority && \
-		go build -o ../conode/build/bcadmin ./byzcoin/bcadmin && \
-		go build -o ../conode/build/csadmin ./calypso/csadmin && \
-		go build -o ../conode/build/phapp ./personhood/phapp
+$(BUILT): build/Dockerfile build/run_nodes.sh
+	make newdb
```
tharvik, line 15: it ensures that sub-`make` are correctly run (right number of concurrent jobs for example).
there is some more to replace below.

```suggestion
	$(MAKE) newdb
```

https://www.gnu.org/software/make/manual/make.html#MAKE-Variable

conode/Makefile
```diff
@@ -1,54 +1,55 @@
...
+	@echo "Starting nodes"
+	docker run -di -v $$(pwd)/build/conodes:/root/conodes --name newdb -p 7770-7777:7770-7777 \
+		$(IMAGE_NAME):dev bash -c \
+		"COTHORITY_ALLOW_INSECURE_ADMIN=true ./run_nodes.sh -n 4 -v 2 -c -d /root/conodes"
+	make docker_wait
+	@echo "Creating new ByzCoin and printing out important variables"
+	docker exec newdb ./bcadmin -c conodes create conodes/public.toml --interval 500ms | tee variables.txt
+	docker exec newdb bash -c './bcadmin key -print $$(ls conodes/key*)' | tee -a variables.txt
+	docker exec newdb bash -c './bcadmin info --bc $$(ls conodes/bc*)' | grep AdminDarc | tee -a variables.txt
+	docker exec newdb chmod -R a+r /root/conodes
```
tharvik, line 30: maybe add a comment

conode/Makefile
```diff
@@ -1,54 +1,55 @@
...
-	@echo "Starting nodes"; \
-	( cd build; COTHORITY_ALLOW_INSECURE_ADMIN=true ./run_nodes.sh -v 2 -t -n 4 -s -d conodes > /dev/null ) & \
-	make docker_wait; \
-	echo "Creating new ByzCoin and printing out important variables"; echo; \
-	rm -f *.cfg; \
-	( ./build/bcadmin -c . create build/conodes/public.toml --interval 500ms; \
-	  ./build/bcadmin latest --bc bc-*; \
-	  ./build/bcadmin key -print key-* ) | egrep "^(ByzCoin|Admin|Private|Creating Spawner)" | tee variables.txt; \
-	echo; echo "Shutting down conodes"; pkill -x conode; rm -f conodes/running
+newdb: docker
```
tharvik, line 19: hum, the "main" generated file in this section is "variables.txt". consider splitting it in two, one `newdb` which `rm build/conode` & `$(MAKE) variables.txt`, one `variables.txt` which does the main work.
it can then be depended upon in `$(WEBAPP_ASSET)`.

conode/Makefile
```diff
@@ -1,54 +1,55 @@
...
+	 echo "AdminDarcID = \"$$( grep AdminDarc variables.txt | sed -e 's/.* //' )\"" >> $@
+	 echo "Ephemeral = \"$$( grep Private variables.txt | sed -e 's/.* //' )\"" >> $@
 	 echo >> $@
 	 cat $(lastword $^) >> $@
 
-docker_run: docker $(WEBAPP_ASSET) docker_start
-
-docker_run_travis: docker
-	docker run --rm -p 7770-7777:7770-7777 --name $(CONTAINER) $(IMAGE_NAME):dev
+docker_run_travis: $(BUILT)
```
tharvik, line 51: it is the same as `docker_start_bg`

conode/Makefile
```diff
@@ -1,54 +1,55 @@
...
-	( cd build; COTHORITY_ALLOW_INSECURE_ADMIN=true ./run_nodes.sh -v 2 -t -n 4 -s -d conodes > /dev/null ) & \
-	make docker_wait; \
-	echo "Creating new ByzCoin and printing out important variables"; echo; \
-	rm -f *.cfg; \
-	( ./build/bcadmin -c . create build/conodes/public.toml --interval 500ms; \
-	  ./build/bcadmin latest --bc bc-*; \
-	  ./build/bcadmin key -print key-* ) | egrep "^(ByzCoin|Admin|Private|Creating Spawner)" | tee variables.txt; \
-	echo; echo "Shutting down conodes"; pkill -x conode; rm -f conodes/running
+newdb: docker
+	rm -f build/conodes/*{db,cfg,key}
```
tharvik, line 20: `$(MAKE) clean`?

conode/variables.txt
```diff
@@ -1,3 +1,7 @@
-ByzCoinID: e9d41ddcaa8433de0d2a85f872a809ff2a53dac19ff82f88cf96f595cd165cb3
-Admin DARC: 843a3a610f4ccb37eaf1188f743c969cbb5698054cdd608b85f5389f0fece783
-Private: dd50f8b8125e7344e7e7f19faeea0821c6a57a595528c4971b7f63a87d30fa0c
+Created ByzCoin with ID 0178ba9de75738c781719770e65da6a1c61b6686f3023148a1bbe7f62a607ffd.
+
+
+export BC="conodes/bc-0178ba9de75738c781719770e65da6a1c61b6686f3023148a1bbe7f62a607ffd.cfg"
+Private: 8431be2e5df68b3bbdcdec5f6e9cd5d10a2987bc5fc819dbcca80c81da400502
+Public: 8afd3d5b99e6d1d803eef091f6acb3dce0ea9167af470649e42d8327b245a3c7
+-- AdminDarc: c03442ce4baf9bf269e0175378b96f43a9e8748f8bdadceeb4c30def17697a4b
```
tharvik, line 7: rm?

webapp/src/assets/config.local.toml
```diff
@@ -1,4 +1,6 @@
-ByzCoinID = "e9d41ddcaa8433de0d2a85f872a809ff2a53dac19ff82f88cf96f595cd165cb3"
+ByzCoinID = "0178ba9de75738c781719770e65da6a1c61b6686f3023148a1bbe7f62a607ffd"
+AdminDarcID = "c03442ce4baf9bf269e0175378b96f43a9e8748f8bdadceeb4c30def17697a4b"
+Ephemeral = "8431be2e5df68b3bbdcdec5f6e9cd5d10a2987bc5fc819dbcca80c81da400502"
```
tharvik, line 3: debug?

.travis.yml
```diff
@@ -18,11 +18,10 @@ jobs:
...
       script:
         - cd conode
-        - export GO111MODULE=on
-        - make docker_run_travis &
+        - make
+        - make docker_run_travis
         - make docker_wait
-        - make utils
-        - build/bcadmin latest --bc bc*cfg
+        - docker exec conode bash -c './bcadmin latest --bc conodes/bc*cfg'
```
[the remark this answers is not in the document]
ineiti: No, that's in fact to check whether it can contact the nodes and get some statistics. Although I'm not sure whether a failing `bcadmin` command would show up as an error here.

conode/Makefile
ineiti: The comment is in the `@echo` - so it appears also during the `make`...

tharvik: a comment for the dev I mean, to know why a `chmod` is necessary

conode/Makefile
ineiti: In fact I couldn't get it to work nicely. Depending on `variables.txt` is a very nice idea, but then splitting it in two is difficult: `docker_start*` then cannot depend on `variables.txt`, but must depend on `newdb`, and so will create the docker everytime again.

But perhaps I'm missing something.

ineiti:
So it's just the comments and the splitting which is missing. If you think it's OK like that, and travis passes, you can merge/rebase it. Else I need more input ;)

tharvik (commented):
> So it's just the comments and the splitting which is missing.

Splitting is mostly done, a bit more consistency and we're good.

> and travis passes, you can merge/rebase it.

LGTM but not for travis :P
```
$ make docker_run_travis
make: *** No rule to make target `docker_run_travis'.  Stop.
```

tharvik (approved):

PR #324, opened 2021-11-23
Add signup and recover possibility

ineiti:
Use the new email part of the personhood service that allows to:
- sign up a new user and send an email with the sign-up link
- recover an existing user with an existing email

### DIFF
--- a/webapp/src/app/newuser/newuser.component.ts
+++ b/webapp/src/app/newuser/newuser.component.ts
@@ -1,7 +1,102 @@
 import { Component, OnInit } from "@angular/core";
+import { FormControl, FormGroup, Validators } from "@angular/forms";
+import { MatSnackBar } from "@angular/material/snack-bar";
+import { MatDialog } from "@angular/material/dialog";
+import PersonhoodRPC from "@dedis/cothority/personhood/personhood-rpc";
+import { ByzCoinService } from "src/app/byz-coin.service";
+import Log from "@dedis/cothority/log";
+import { showDialogInfo, showTransactions, TProgress } from "src/lib/Ui";
+import { ServerIdentity } from "@dedis/cothority/network";
 
 @Component({
-  selector: "app-newuser",
-  templateUrl: "./newuser.component.html",
+    selector: "app-newuser",
+    templateUrl: "./newuser.component.html",
 })
-export class NewuserComponent {}
+export class NewuserComponent implements OnInit {
+    signupForm: FormGroup;
+    recoveryForm: FormGroup;
+
+    constructor(private snack: MatSnackBar,
+                private dialog: MatDialog,
+                public bcs: ByzCoinService) {
+    }
+
+    async ngOnInit() {
+        for (const node of this.bcs.config.roster.list) {
+            Log.info("node", node.description, "id", node.id);
+        }
+        this.signupForm = new FormGroup({
+            alias: new FormControl(),
+            email: new FormControl("", Validators.email),
+        });
+        this.recoveryForm = new FormGroup({
+            email: new FormControl("", Validators.email),
+        });
+    }
+
+    getPHRPC(): PersonhoodRPC {
+        if (this.bcs.config.signupNode === "") {
+            throw new Error("didn't find signup node");
+        }
+        const node = new ServerIdentity({
+            url: this.bcs.config.signupNode,
+            id: Buffer.alloc(32)
+        });
+        return new PersonhoodRPC(node);
+    }
+
+    async signup() {
+        const reply = await showTransactions(this.dialog, "Signing up new user",
+            async (progress: TProgress) => {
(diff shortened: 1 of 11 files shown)


webapp/src/app/newuser/newuser.component.ts
```diff
@@ -1,7 +1,102 @@
...
-  templateUrl: "./newuser.component.html",
+    selector: "app-newuser",
+    templateUrl: "./newuser.component.html",
 })
-export class NewuserComponent {}
+export class NewuserComponent implements OnInit {
+    signupForm: FormGroup;
+    recoveryForm: FormGroup;
+
+    constructor(private snack: MatSnackBar,
```
tharvik, line 19: `snack` is unused

webapp/src/app/newuser/newuser.component.ts
```diff
@@ -1,7 +1,102 @@
...
+                private dialog: MatDialog,
+                public bcs: ByzCoinService) {
+    }
+
+    async ngOnInit() {
+        for (const node of this.bcs.config.roster.list) {
+            Log.info("node", node.description, "id", node.id);
+        }
```
tharvik, line 27: debug?

webapp/src/app/newuser/newuser.component.ts
```diff
@@ -1,7 +1,102 @@
...
+        this.signupForm = new FormGroup({
+            alias: new FormControl(),
+            email: new FormControl("", Validators.email),
+        });
+        this.recoveryForm = new FormGroup({
+            email: new FormControl("", Validators.email),
+        });
```
tharvik, line 34: you can move theses to the constructor and have the fields readonly

webapp/src/assets/config.dedis.toml
```diff
@@ -1,4 +1,5 @@
 ByzCoinID = "9cc36071ccb902a1de7e0d21a2c176d73894b1cf88ae4cc2ba4c95cd76f474f3"
+SignupNode = "https://byzcoin-signup.c4dt.org"
```
tharvik, line 2: not really related to the PR; `curl http://byzcoin-signup.c4dt.org` (via http not https) gives me a 404, shouldn't it redirect to the https version instead?

webapp/src/lib/config.ts
```diff
@@ -1,6 +1,7 @@
 import { Roster } from "@dedis/cothority/network/proto";
 
 import toml from "toml";
+import Log from "@dedis/cothority/log";
```
tharvik, line 4: unused

webapp/src/lib/config.ts
```diff
@@ -26,7 +27,17 @@ export class Config {
...
                 throw Error("is not a string");
             }
-            if (!(/[a-f0-9]{64}/).test(field)) {
+            if (!(/^[a-f0-9]{64}$/).test(field)) {
+                throw Error("is not of correct format");
+            }
+            return Buffer.from(field, "hex");
+        };
+
+        const asUUID = (field: any): ID => {
```
tharvik, line 36: unused

webapp/src/app/newuser/newuser.component.ts
```diff
@@ -1,7 +1,102 @@
...
+    }
+
+    getPHRPC(): PersonhoodRPC {
+        if (this.bcs.config.signupNode === "") {
```
tharvik, line 38: the field is `undefined` if it is not existing in the config, not the empty string

```suggestion
        if (this.bcs.config.signupNode === undefined) {
```

webapp/src/lib/config.ts
```diff
@@ -43,6 +54,7 @@ export class Config {
         return new Config(
             getField("ByzCoinID", asID),
             Roster.fromTOML(raw),
+            tryToGetField("SignupNode", asString),
```
tharvik, line 57: consider parsing it as URL to ensure that it is in the correct format and fails at load time, not latter

tharvik (changes requested):
some cosmetics (I looove `private readonly`) and a bug when missing the `signupNode` from the config

webapp/src/lib/config.ts
ineiti: removed

webapp/src/lib/config.ts
ineiti: Added `asURL`, even though for the `ServerIdentity` I need to convert it back to a string...

tharvik: yep, ideally, `ServerIdentity` should take a URL, but for now, at least the configuration will be correct.

I remember that some times ago, I already wanted to change cothority to use the correct URL type, to catch misconfiguration early, but it was cumbersome as it wasn't working well with protobuf.

tharvik (approved):

PR #326, opened 2021-11-24
Deploy on main

ineiti:
It didn't deploy with the error

```
Skipping a deployment with the script provider because this branch is not permitted: main
```

So I guess it needs to be on the same line...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,9 +71,8 @@ jobs:
       deploy:
         provider: script
         script: ssh omniledger@c4dtsrv1.epfl.ch bin/update.sh
-      on:
-        branch:
-          main
+        on:
+          branch: main
 
     - name: NPM
       language: node_js


.travis.yml
```diff
@@ -72,8 +72,7 @@ jobs:
         provider: script
         script: ssh omniledger@c4dtsrv1.epfl.ch bin/update.sh
       on:
-        branch:
-          main
+        branch: main
```
cgrigis, line 75: I'm surprised, because those two should be equivalent.

Perhaps:
```
    deploy:
      ...
      on:
        branch: main
```
i.e. `on:` **indented** under `deploy:`?

ineiti: Thanks! I think that's it. I _did_ think that it's strange it didn't work that way. But indenting `on` under `deploy` looks definitely better!

https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on

I'm not sure if I should blame yaml for this, or be happy that yaml is so concise...

cgrigis (approved):

PR #327, opened 2021-11-25
Correctly recover user

ineiti:
Currently when recovering a user, it tries to:
1. update the signer-darc of the user
2. insert the device in the list of the user

But as of #263, recoveries are only allowed to evolve the user-signer-darc,
but not to sign on its behalf. So the two steps need to be signed by
different signers:
1. can be signed by the user that does the recovery
2. must be signed by the new ephemeral key

### DIFF
--- a/webapp/src/app/admin/contacts/contacts.component.ts
+++ b/webapp/src/app/admin/contacts/contacts.component.ts
@@ -210,11 +210,11 @@ export class ContactsComponent implements OnInit {
                         this.user.credStructBS.credPublic.alias.getValue(),
                         now.getFullYear(), now.getMonth() + 1, now.getDate(), now.getHours(), now.getMinutes());
                     const ephemeralIdentity = SignerEd25519.random();
-                    let newDarc: Darc;
-                    await this.user.executeTransactions((tx) => {
-                        newDarc = tx.spawnDarcBasic(`recover:${name}`, [ephemeralIdentity]);
+                    const newDarc = await this.user.executeTransactions((tx) => {
+                        const d = tx.spawnDarcBasic(`recover:${name}`, [ephemeralIdentity]);
                         const idEph = newDarc.getBaseID();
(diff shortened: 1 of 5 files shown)


webapp/src/app/admin/contacts/contacts.component.ts
```diff
@@ -204,15 +204,23 @@ export class ContactsComponent implements OnInit {
...
                     }
 
-                    progress(70, "Creating new device for recovery");
+                    progress(50, "Creating new device for recovery");
                     const now = new Date();
                     const name = sprintf("recovered by %s at %d/%02d/%02d %02d:%02d",
                         this.user.credStructBS.credPublic.alias.getValue(),
                         now.getFullYear(), now.getMonth() + 1, now.getDate(), now.getHours(), now.getMinutes());
                     const ephemeralIdentity = SignerEd25519.random();
+                    let newDarc: Darc;
```
tharvik, line 213: Hum, it's weird for me to mutate variable outside a closure, as it might get run latter, or multiple times, one doesn't know from a quick look.

Just an idea, see how it fits: `this.user.executeTransactions` can be updated to be
```ts
async executeTransactions<T>(
    addTxs: (tx: SpawnerTransactionBuilder) => Promise<T> | T,
    wait = 0
): Promise<T>
```
which allows to have `const newDarc = await this.user.executeTransactions` with a closure returning the darc.

This way, you avoid the closure potentially modify `newDarc` unexpectedly.

tharvik (approved):

webapp/src/app/admin/contacts/contacts.component.ts
ineiti: I had the same thoughts, but does your proposition work? I thought I'd also have to change the `SpawnerTransactionBuilder`, so I didn't do it, rabbithole and everything.

ineiti: In fact it's enough - thanks, that's much cleaner.

tharvik (approved):

PR #332, opened 2021-12-20
Move to GitHub actions

tharvik:
Drop travis, use GA


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
+name: deploy
+
+on: [pull_request]
```
ineiti, line 3: Shouldn't this be on `main` only?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+
+env:
+    DEPLOY_USER: omniledger
+    DEPLOY_HOST: srv1.c4dt.org
+
+jobs:
+    deploy-website:
```
ineiti, line 10: Remind me how it is assured that this is only run if the tests pass?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,50 @@
...
+            - uses: actions/checkout@v2
+
+            - uses: actions/setup-node@v1
+              with:
+                  node-version: 12
+                  cache: npm
+
+            - run: npm install
+
+            - run: npm test
```
ineiti, line 42: I'm confused - shouldn't the test be run _before_ all else?

ineiti (changes requested):
Perhaps I misunderstood the workflow you wrote. But why didn't you add a `test` workflow that runs on every pull-request, and a `deploy` that runs only when merged on `main`?

tharvik:
> Perhaps I misunderstood the workflow you wrote. But why didn't you add a `test` workflow that runs on every pull-request, and a `deploy` that runs only when merged on `main`?

It wasn't ready for review. I had to try out the deploy stage so I "forced" triggered it, via a "TMP force deploy" commit.

.github/workflows/deploy.yml
tharvik: The `if: ${{ github.event.workflow_run.conclusion == 'success' }}` and the deps on another workflow ensure that.

.github/workflows/deploy.yml
tharvik: It's ran in the "lint and test" workflow

ineiti:
> It wasn't ready for review. I had to try out the deploy stage so I "forced" triggered it, via a "TMP force deploy" commit.

Ah, sorry. I saw the "review requested" in the github-status...

@cgrigis showed me the https://github.com/c4dt/byzcoin/blob/master/.github/workflows/update.yaml#L6

Then you can manually trigger the commit, too:

https://github.com/c4dt/byzcoin/actions/workflows/update.yaml

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,114 @@
...
+        name: build conode
+        runs-on: ubuntu-latest
+        defaults:
+            run:
+                working-directory: conode
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-go@v2
+
+            - run: make docker_start_detached docker_wait
```
ineiti, line 16: Why does this use two `-run:`? In the other actions there is a
```
run: |
  make...
  docker...
```

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,114 @@
...
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
+
+            - uses: actions/checkout@v2
+              with:
+                  repository: dedis/cothority
+                  path: cothority
+            - run: make docker
+              working-directory: cothority
```
ineiti, line 34:
```suggestion
            - run: ( cd cothority && make docker )
```

Or is that too cryptic?

Again: why do you prefer to have multiple `-run:` here instead of a `-run: |`?

Same question for the other steps.

.github/workflows/lint-and-test.yml
```diff
@@ -0,0 +1,114 @@
...
+        runs-on: ubuntu-latest
+        needs: lint-webapp
+        defaults:
+            run:
+                working-directory: webapp
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
+
+            - run: npm install
```
ineiti, line 64: Shouldn't we use `npm ci`? I think I had bad experiences with `npm i`, which does also update the libraries. And I'm not sure the `package.json` is correctly written to survive `npm i` also in the future...

webapp/cas/go.mod
```diff
@@ -13,3 +13,37 @@ require (
 	go.dedis.ch/onet/v3 v3.2.10
 	go.dedis.ch/protobuf v1.0.11
 )
+
+require (
```
ineiti, line 17: Lazy question: is this the new `go.mod` format with two `require` sections? One for the requirements, and one for the indirect ones?

ineiti (changes requested):
Some questions in the workflow files.

Also, my head parses github-workflow files poorly, so I cannot tell if the `on:...` are correct and will do what is expected from them. We'll see on the next PR :)

ineiti:
Oh, and you'll also need to remove the travis requirements in github, else it will always show `14 successful and 2 expected`.

.github/workflows/lint-and-test.yml
tharvik: If find it a bit more readable, especially when reading the resulting build log.

.github/workflows/lint-and-test.yml
tharvik: Ho, I though it was only a quick way to remove the node_modules and install the pkgs, but [the documentation] states that it is designed for CI and behave differently.

webapp/cas/go.mod
tharvik: Dunno, golang 1.17 forces me to add the indirect deps and generated this format. But moving indirect deps to direct doesn't makes `go mod tidy` reorder theses.

.github/workflows/lint-and-test.yml
tharvik: IMO, using `working-directory` makes the job more readable, by removing the `cd dynacred &&` for each step.

tharvik:
> Also, my head parses github-workflow files poorly, so I cannot tell if the `on:...` are correct and will do what is expected from them. We'll see on the next PR :)

Yeah, I'm never sure either, let's hope (and be ready for a hotfix!)

> you'll also need to remove the travis requirements in github

It seems that I can only remove travis from the whole organisation, not from a single repo. Keeping it for now to ensure that we don't break some other CI.

ineiti:
I meant here:

https://github.com/c4dt/omniledger/settings/branch_protection_rules/6643219

Remove the two travis checks, and add the 12 github-workflows.

ineiti (approved):

PR #333, opened 2021-12-22
github/deploy: build locally

### LINKED ISSUE #334: npm build --prod fails
When building the webapp with `npm build --prod`, the resulting scripts do nothing: when loading the `index.html` there is no output in the console.

Currently only `npm build` works. One could probably use `git bisect`, or do it manually, to find the commit that makes it fail.

I guess the following might be the culprit, as it updates to the latest cothority:

https://github.com/c4dt/omniledger/commit/1f694e281d410309eb78818efbf02b6b05d53853

tharvik:
* only deploy on main branch (and don't run tests)
* build webapp in CI
* avoid building for old browsers
* fix `ng build --prod`

Closes #334

### DIFF
--- a/.github/workflows/lint-and-test.yml
+++ b/.github/workflows/lint-and-test.yml
@@ -1,6 +1,6 @@
 name: lint and test
 
-on: [push, pull_request]
+on: pull_request
 
 jobs:
     build-conode:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -13,18 +13,45 @@ jobs:
         name: deploy website
         runs-on: ubuntu-latest
         if: ${{ github.event.workflow_run.conclusion == 'success' }}
+        defaults:
+            run:
+                working-directory: webapp
         steps:
             - uses: actions/checkout@v2
+
+            - uses: actions/setup-node@v2
+              with:
+                  node-version: 12
+                  cache: npm
+                  cache-dependency-path: webapp/package-lock.json
 
             - uses: kielabokkie/ssh-key-and-known-hosts-action@v1.2.0
               with:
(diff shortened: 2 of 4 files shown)


ineiti:
> Sadly, I was unable to push a working build, maybe you have an idea @ineiti ?

What is the problem exactly? What do you mean by "push a working build" - is this about the deployment to the server?

tharvik:
Yep, I tried inlining the `bin/update.sh` script, to ensure that it is correctly built before actually deploying it. But when forcing to deploy, the resulting website doesn't do anything: it loades the assets but doesn't seems continue with application launch.

ineiti:
OK - that is the problem I have with `npm build --prod` :( This is why I only do a `npm build` now, until I can figure out why the `--prod` fails. I'm not really sure which change broke this. It might be the wrong node-version or a library that has been updated in the meantime. Probably a `git bisect` on the `omniledger` repo could show when it fails.

A possible failing commit is at #334

tharvik:
I found a fix, disabling "buildOptimizer" in angular.json correctly deployes it. It is currently deploy on https://login.c4dt.org and https://demo.c4dt.org/omniledger

.github/workflows/lint-and-test.yml
```diff
@@ -1,6 +1,6 @@
 name: lint and test
 
-on: [push, pull_request]
+on: [pull_request]
```
ineiti, line 3:
```suggestion
on: pull_request
```

ineiti (dismissed):
Great catch for the buildOptimizer!

ineiti:
I added a `Closes #334` to the description.

ineiti (approved):

PR #340, opened 2022-02-17
Darc identity tsm

ineiti:
Makes the `Calypso` and `TransactionBuilder` a bit more fine-grained so that the TSM signatures can also be used. The main difference with the other signatures is that TSM signatures are asynchronous.

Also includes some fixes to make it possible to use `npm link @deds/cothority` and `npm link @dedis/kyber` without hell breaking loose.


pm-duokey:
Addition of a threshold security module signer from [Sepior](https://sepior.com/).

PR #347, opened 2022-03-04
Updating addresses of c4dt-conode servers

ineiti:


tharvik:
After discussion, find a ground truth for roster configs between ansible and here.

PR #350, opened 2022-03-18
Correctly launch two commands with ssh

ineiti:
@tharvik 's always right...


tharvik (approved):
> @tharvik 's always right...

And I'm always lying :smiling_imp:

ineiti:
> And I'm always lying 😈

https://www.smbc-comics.com/comic/on-the-island

PR #368, opened 2022-11-07
cas/matrix: handle double SSO

tharvik:
Now that we have two SSO, we need to actively select CAS.


ineiti (approved):
Not relevant anymore here, but I think generally `go.sum` should be as a binary, then it doesn't appear in PRs...

tharvik:
> I think generally `go.sum` should be as a binary, then it doesn't appear in PRs...

but it's not a binary, it is text (:

I'm using the `-diff` attribute to disallow even showing a diff, but yeah, quite the same thing

PR #383, opened 2023-09-21
Archiving Omniledger

ineiti:
As users will use the signup link to connect, make it so it doesn't break.


PascalinDe:
sign-up works locally with the Byzcoin setup so LGTM

PascalinDe (approved):

